PR #7, opened 2017-08-23
GCE Support for GCP Auth Backend 

emilymye:


jefferai:
I'm not a huge fan of mixing and matching two JOSE libraries in the same code files. I assume that the go-jose dep is imposed by some of the new google libraries? If so, what are your thoughts on just migrating the SermoDigital/jose bits to go-jose?

emilymye:
@jefferai re:jose i think the pluginutil in hashicorp vault uses SermoDigital/jose so idk if it's plausible to convert all of Vault over.  I actually added this dep because I wanted JWK support.

jefferai:
I didn't mean converting all of Vault, I just meant using one or the other in your plugin, it feels strange to mix and match two libraries that do the same thing in the same source file.

Just curious, what are the JWKs being used for? I don't see those in the wild much.

jefferai:
I have no issue with the project having two libraries due to deps, I'm talking specifically about https://github.com/hashicorp/vault-plugin-auth-gcp/pull/7/files#diff-7208fb653f9b4fd9a648f0c759329731

emilymye:
Oh then, sure! I'll go ahead and refactor

plugin/backend.go
```diff
@@ -104,6 +108,28 @@ func (b *GcpAuthBackend) IAM(s logical.Storage) (*iam.Service, error) {
 	return b.iamClient, nil
 }
 
+func (b *GcpAuthBackend) GCE(s logical.Storage) (*compute.Service, error) {
+	b.clientMutex.RLock()
```
jefferai: Is there a chance that you'll have the same mount being used for GCE and IAM? Or is it strictly one or the other?

jefferai: OK, I see the answer is yes. It might make things more complicated, might not, but may be worth splitting out the mutex so it's one per type. But I doubt there will be contention issues if not, so feel free not to!

plugin/path_login.go
```diff
@@ -132,91 +134,141 @@ func (b *GcpAuthBackend) parseInfoFromJwt(req *logical.Request, data *framework.
...
+			return nil, err
+		}
+
+		accountKey, err := util.ServiceAccountKey(iamClient, keyId, serviceAccountId, role.ProjectId)
+		if err != nil {
+			return nil, err
+		}
+
+		return util.PublicKey(accountKey.PublicKeyData)
+	default:
```
jefferai: Minor nitpick: it's usually nice to explicitly name your cases (if you have them) and have the default be a "this should never happen" error condition. Mostly it makes sure that if things change later the default behavior doesn't accidentally come back to bite you.  :-)

plugin/path_login.go
```diff
@@ -132,91 +134,141 @@ func (b *GcpAuthBackend) parseInfoFromJwt(req *logical.Request, data *framework.
...
+		var certsEndpoint string
+		conf, err := b.config(s)
+		if err != nil {
+			return nil, fmt.Errorf("could not read config for backend: %v", err)
+		}
+		if conf != nil {
+			certsEndpoint = conf.GoogleCertsEndpoint
+		}
+
+		key, err := util.Oauth2RSAPublicKey(keyId, certsEndpoint)
```
jefferai: Another minor nitpick: as long as you're camelcasing/properly capitalizing, note that it's OAuth, not Oauth :-)

plugin/path_login.go
```diff
@@ -298,6 +352,187 @@ func (b *GcpAuthBackend) validateAgainstIAMRole(serviceAccount *iam.ServiceAccou
...
+			zone = instance.Zone[idx:len(instance.Zone)]
+		} else {
+			// Expect full zone name to be set as instance zone.
+			zone = instance.Zone
+		}
+
+		if zone != role.Zone {
+			return fmt.Errorf("instance is not in role zone '%s'", role.Zone)
+		}
+	} else if len(role.Region) > 0 {
```
jefferai: Is this the proper behavior if both role.Zone and role.Region are set, or should these be independent if clauses?

plugin/path_login.go
```diff
@@ -298,6 +352,187 @@ func (b *GcpAuthBackend) validateAgainstIAMRole(serviceAccount *iam.ServiceAccou
...
+		var group *compute.InstanceGroup
+		var err error
+
+		// Check if group should be zonal or regional.
+		if len(role.Zone) > 0 {
+			group, err = gceClient.InstanceGroups.Get(role.ProjectId, role.Zone, role.InstanceGroup).Do()
+			if err != nil {
+				return fmt.Errorf("could not find role instance group %s (project %s, zone %s)", role.InstanceGroup, role.ProjectId, role.Zone)
+			}
+		} else if len(role.Region) > 0 {
```
jefferai: Same question.

plugin/path_role.go
```diff
@@ -12,71 +13,133 @@ import (
...
 )
 
+var baseRoleFieldSchema map[string]*framework.FieldSchema = map[string]*framework.FieldSchema{
+	"name": {
+		Type:        framework.TypeString,
+		Description: "Name of the role.",
+	},
+	"type": {
+		Type:        framework.TypeString,
+		Description: "Type of the role. Currently supported: iam",
```
jefferai: Shouldn't this be "iam, gce"?

plugin/path_role.go
```diff
@@ -12,71 +13,133 @@ import (
...
+	"ttl": {
+		Type:    framework.TypeDurationSecond,
+		Default: 0,
+		Description: `
+	Duration in seconds after which the issued token should expire. Defaults to 0,
+	in which case the value will fallback to the system/mount defaults.`,
+	},
+	"max_ttl": {
+		Type:        framework.TypeDurationSecond,
+		Default:     0,
```
jefferai: Putting default 0 is fine, but also not necessary :-)

plugin/path_role.go
```diff
@@ -12,71 +13,133 @@ import (
...
+var baseRoleFieldSchema map[string]*framework.FieldSchema = map[string]*framework.FieldSchema{
+	"name": {
+		Type:        framework.TypeString,
+		Description: "Name of the role.",
+	},
+	"type": {
+		Type:        framework.TypeString,
+		Description: "Type of the role. Currently supported: iam",
+	},
+	"policies": {
```
jefferai: This should be a TypeCommaStringSlice, which lets people specify it as either a "comma,separated,string" or via JSON as a ["json", "array"] !

plugin/path_role.go
```diff
@@ -12,71 +13,133 @@ import (
...
+	If set, indicates that the token generated using this role should never expire. The token should be renewed within the
+	duration specified by this value. At each renewal, the token's TTL will be set to the value of this parameter.`,
+	},
+	// -- GCP Information
+	"project_id": {
+		Type:        framework.TypeString,
+		Description: `The id of the project that authorized instances must belong to for this role.`,
+	},
+	"service_accounts": {
+		Type: framework.TypeString,
```
jefferai: Same here, should be TypeCommaStringSlice.

plugin/path_role.go
```diff
@@ -107,9 +170,57 @@ is assumed to be all service accounts under the role's project.`,
...
+				},
+			},
+			Callbacks: map[logical.Operation]framework.OperationFunc{
+				logical.UpdateOperation: b.pathRoleEditIamServiceAccounts,
+			},
+			HelpSynopsis:    "Add or remove service accounts for an existing `iam` role",
+			HelpDescription: "Add or remove service accounts from the list bound to an existing `iam` role",
+		},
+
+		// Edit service accounts on an GCE role
```
jefferai: Comment should say "labels" not "service accounts"

plugin/path_role.go
```diff
@@ -318,20 +427,34 @@ func (b *GcpAuthBackend) role(s logical.Storage, name string) (*gcpRole, error)
...
-	return nil, s.Put(entry)
+	if err := s.Put(entry); err != nil {
+		return nil, err
+	}
+
+	return resp, nil
 }
 
 type gcpRole struct {
+	// TODO(emilyyme): Make this consistent with input arg name ("type").
```
jefferai: It's not super important though, and be careful about backwards compatibility if you want to change it. e.g. it's fine to change the name of the actual struct member but changing the struct tags can lead to danger without an upgrade path.

plugin/path_role.go
```diff
@@ -444,68 +555,196 @@ func (role *gcpRole) updateRole(sys logical.SystemView, op logical.Operation, da
...
+	if ok {
+		role.AllowGCEInference = allowGCEInference.(bool)
+	} else if op == logical.CreateOperation {
+		role.AllowGCEInference = data.Get("allow_gce_inference").(bool)
+	}
+
+	maxJwtExp, ok := data.GetOk("max_jwt_exp")
+	if ok {
+		role.MaxJwtExp = time.Duration(maxJwtExp.(int)) * time.Second
+	} else {
```
jefferai: Did you mean this to be `else if op == logical.CreateOperation`? As is, anytime someone updates but doesn't specify this field it will reset the value.

plugin/backend.go
emilymye: I'll keep it as is and if someone actually runs into an issue with it, we can change it :)

plugin/path_login.go
emilymye: I keep doing that! Will fix

plugin/path_role.go
emilymye: Yeah, pretty much why I put a TODO - I wasn't really sure the best way to migrate it. I can remove it if you think it's not worth the effort.

plugin/path_login.go
emilymye: done

plugin/path_login.go
emilymye: If Zone is set, we will ignore Region and use the region of the given Zone.

plugin/path_role.go
emilymye: whoops, fixed

plugin/path_role.go
emilymye: done

plugin/path_role.go
emilymye: done

plugin/path_role.go
emilymye: done

plugin/path_role.go
emilymye: sg

plugin/path_role.go
emilymye: Yes!

plugin/path_role.go
jefferai: Weird. I'll have to look into that. It ought to work.

jefferai:
Hi Emily,

From my end things look really good. I do want to stress that I have not tested functionality -- I'm leaving that to you and your immediately available resources :-D  But from a code quality/Vault standards perspective I think it's good.

One thing that I'll request but not insist on -- in the aws auth backend you'll notice that the items in roles that are specifically for selecting users/accounts/values (subnet, vpc, etc.) are prefixed with `bound_` to make it clear that it's a binding requirement, e.g. `bound_ami_id`. We are also doing this with the Kubernetes auth plugin that is currently under development. I think it'd be nice if this plugin also used similar terminology, but if you have strong feelings against then it's okay.

emilymye:
Hi Jeff,

I don't have objections to that other than the fact that we aren't using "bound_" for some of the IAM attributes - do you think it would be a big deal to change those parameter names?

jefferai:
You can basically add the bound_ version and do a check for that version; if it's empty or not set, check for the old version. Then you just document the bound_ version.

emilymye:
This is ready to go for the most part (I figured out the issue we talked about on chat). I need to add some tests but is there anything else you wanted me to address?

jefferai:
Hi Emily,

Browsed through your most recent commits -- looks good to me! Fix the conflicts and I'll merge it in!

emilymye:
Rebased! I'll add the docs PR to vault main soon but this should be ready to merge.

jefferai:
:clickety:

PR #8, opened 2017-08-31
Update README with note about plugin error when missing env variables

emilymye:
Fleetsmith ran into errors where the server failed to mount a custom plugin with TLS enabled because their script hadn't set VAULT_CAPATH. Server startup and plugin catalog both succeeded but the auth-enable step using the plugin backend failed with a non-descript error. Will file an issue against Vault main as well for better errors, but we wanted to document this issue in case.

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,7 +24,7 @@ Otherwise, first read this guide on how to [get started with Vault](https://www.
 
 To learn specifically about how plugins work, see documentation on [Vault plugins](https://www.vaultproject.io/docs/internals/plugins.html).
 
-### Usage
+## Usage
 
 Please see [documentation for the plugin](https://www.vaultproject.io/docs/auth/gcp.html)
 on the Vault website.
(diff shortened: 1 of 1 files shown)


calvn:
The empty map on the error field is due to `hclog` (more specifically the stdlib JSON marshaler) not marshaling error types. The plugin process returns an empty marshaled error value to the plugin client (that's running in Vault). The error is basically getting marshaled to an empty `{}` JSON object that is sent to the client, so when the logger on the client side parses the output it's a `[]map` on the error field. 

References:
https://github.com/hashicorp/go-hclog/pull/9
https://github.com/golang/go/issues/10748

calvn:
I believe the CA path can be provided when the plugin is being registered to the catalog:

```
$ vault write sys/plugins/catalog/my-plugin sha256=<sha256> command='some-plugin -ca-path=</path/to/ca>'
```

jefferai:
Just a note that the proposed fix for this is waiting on some review by people that are currently out.

jefferai:
Just a note that the go-hclog pull request got merged in. I'll update the dep in this repo; I think that will also then close this PR?

PR #12, opened 2017-09-26
Add GCE metadata to inferred IAM auth token.

emilymye:
Adding so that a) inferred tokens can be rejected for renewal if the role changes to not allow GCE inferrence, and b) to audit GCE instances better.

Also a small fix for an error message in iam utils.

### DIFF
--- a/plugin/path_login.go
+++ b/plugin/path_login.go
@@ -263,7 +263,7 @@ func (b *GcpAuthBackend) pathIamLogin(req *logical.Request, loginInfo *gcpLoginI
 	role := loginInfo.Role
 	if !role.AllowGCEInference && loginInfo.GceMetadata != nil {
 		return logical.ErrorResponse(fmt.Sprintf(
-			"IAM role '%s' does not allow gce inference but GCE instance metadata token given", loginInfo.RoleName)), nil
+			"Got GCE token but IAM role '%s' does not allow GCE inference", loginInfo.RoleName)), nil
 	}
 
 	// TODO(emilymye): move to general JWT validation once custom expiry is supported for other JWT types.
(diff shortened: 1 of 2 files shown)


plugin/path_login.go
```diff
@@ -399,24 +405,27 @@ func (b *GcpAuthBackend) pathGceLogin(req *logical.Request, loginInfo *gcpLoginI
 		return logical.ErrorResponse(err.Error()), nil
 	}
 
+	iamClient, err := b.IAM(req.Storage)
```
vishalnayak: Can we remove `fmt.Printf()` from lines 379 and 380?

plugin/path_login.go
```diff
@@ -428,6 +437,25 @@ func (b *GcpAuthBackend) pathGceLogin(req *logical.Request, loginInfo *gcpLoginI
 	return resp, nil
 }
 
+func authMetadata(loginInfo *gcpLoginInfo, serviceAccount *iam.ServiceAccount) map[string]string {
```
vishalnayak: Not that I see a problem currently, but keep in mind that all the items here will be audit logged in plaintext.

plugin/path_login.go
```diff
@@ -326,6 +322,11 @@ func (b *GcpAuthBackend) pathIamRenew(req *logical.Request, role *gcpRole) error
 		return fmt.Errorf(clientErrorTemplate, "IAM", err)
 	}
 
+	roleName, ok := req.Auth.Metadata["role"]
```
vishalnayak: Will this not block the renewals of already issued tokens?

emilymye: Technically, this check is already done beforehand, so it shouldn't. I can just change it to a parameter.

plugin/path_login.go
emilymye: 👍

plugin/path_login.go
emilymye: Yup!

jefferai:
LGTM. Vishal, if you're happy with the changes, feel free to merge.

PR #20, opened 2017-12-04
Update deps

emilymye:


jefferai:
Any reason you switched the Vault dep from master to 0.9.0?

emilymye:
Not sure since I just ran the dep tool but I'll fix it.

emilymye:
This should be ready to go

jefferai:
Cool -- I don't normally check dep updates carefully but I do look at changes to the toml file and that struck me as odd. I may at release time change things to a specific branch for reasons of pulling-into-vault's-codebase-for-compiling-in but since release always comes from master generally that'll be the right thing at the time anyways :-)

PR #21, opened 2017-12-04
Fix oauth key for GCE inferrence (#19)

### LINKED ISSUE #19: IAM roles with allow_gce_inference fail on GCE tokens
### Issue

When attempting to use a GCE token retrieved from the metadata server the IAM role login endpoint responds with a 404 indicating that it cannot find the service account key for the inferred service account. This is because it is attempting to look up a service account key using the `kid` field from the JWT header, but because this is a GCE token the `kid` field is actually one of Google's OAuth keys.

### Suggested Fix

The following line checks for the service account key regardless of the token type:
https://github.com/hashicorp/vault-plugin-auth-gcp/blob/4febf0d5513b41e1a3b46036e8e2d03469235205/plugin/path_login.go#L207

Instead there should probably be a conditional for inferred tokens that instead does the exact same thing as the following `case` statement:
https://github.com/hashicorp/vault-plugin-auth-gcp/blob/4febf0d5513b41e1a3b46036e8e2d03469235205/plugin/path_login.go#L213-L227

### Reproduction steps
Assuming you have an instance with an associated service account named `vault-client`:

```sh
$ vault auth-enable gcp
$ vault write auth/gcp/role/<your project here> type="iam" project_id="<your project here>" bound_service_accounts="*" policies="default" allow_gce_inference="true"
$ GOOGLE_PROJECT=$(gcloud config get-value project)
$ SERVICE_ACCOUNT=vault-client@${GOOGLE_PROJECT}.iam.gserviceaccount.com
$ ROLE=$GOOGLE_PROJECT
$ curl -H "Metadata-Flavor: Google" -G --data-urlencode "audience=$VAULT_ADDR/vault/$ROLE" --data-urlencode "format=full" "http://metadata/computeMetadata/v1/instance/service-accounts/$SERVICE_ACCOUNT/identity" > token.jwt
$ vault write auth/gcp/login role="$ROLE" jwt=@token.jwt
```

Result:
```
Error writing data to auth/gcp/login: Error making API request.

URL: PUT http://<redacted>.internal:8200/v1/auth/gcp/login
Code: 400. Errors:

* service account key 'projects/<redacted>/serviceAccounts/110265710746254174717/keys/614d0ed93d39fbf1bc184793da0801d40f40b822' does not exist: googleapi: Error 404: Service account key projects/<redacted>/serviceAccounts/110265710746254174717/keys/614d0ed93d39fbf1bc184793da0801d40f40b822 does not exist., notFound
```

Thanks!

emilymye:
Fixes #19

### DIFF
--- a/plugin/path_login.go
+++ b/plugin/path_login.go
@@ -206,28 +206,35 @@ func (b *GcpAuthBackend) getSigningKey(token *jwt.JSONWebToken, rawToken string,
 
 		accountKey, err := util.ServiceAccountKey(iamClient, keyId, serviceAccountId, role.ProjectId)
 		if err != nil {
-			return nil, err
+			// Attempt to get a normal Google Oauth cert in case of GCE inferrence.
+			key, err := b.getGoogleOauthCert(keyId, s)
+			if err != nil {
+				return nil, errors.New("could not find service account key or Google Oauth cert with given 'kid' id")
+			}
 		}
-
 		return util.PublicKey(accountKey.PublicKeyData)
 	case gceRoleType:
-		var certsEndpoint string
-		conf, err := b.config(s)
-		if err != nil {
-			return nil, fmt.Errorf("could not read config for backend: %v", err)
-		}
-		if conf != nil {
-			certsEndpoint = conf.GoogleCertsEndpoint
-		}
-
-		key, err := util.OAuth2RSAPublicKey(keyId, certsEndpoint)
-		if err != nil {
-			return nil, err
-		}
-		return key, nil
+		return b.getGoogleOauthCert(keyId, s)
 	default:
 		return nil, fmt.Errorf("unexpected role type %s", role.RoleType)
 	}
+}
+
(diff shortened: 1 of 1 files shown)


plugin/path_login.go
```diff
@@ -206,30 +206,37 @@ func (b *GcpAuthBackend) getSigningKey(token *jwt.JSONWebToken, rawToken string,
 
 		accountKey, err := util.ServiceAccountKey(iamClient, keyId, serviceAccountId, role.ProjectId)
 		if err != nil {
-			return nil, err
+			// Attempt to get a normal Google Oauth cert in case of GCE inferrence.
+			key, err := b.getGoogleOauthCert(keyId, s)
```
jefferai: Do you need to do something with the returned key here?

emilymye: fixed

jefferai:
@emilymye good to merge on your end?

emilymye:
sgtm!

PR #23, opened 2017-12-22
Add CLI helper for generating IAM JWT.

emilymye:


jefferai:
Hi @emilymye !

I'm going to sort of lean on "assuming you've tested it and it works" here. One thing I would suggest you do is look at the new code under `command` and with the other auth helpers. I'm not sure if the API has changed per se, but there may be easier ways to deal with e.g. formatting.

jefferai:
I've updated deps in the repo.

emilymye:
@jefferai  I updated the CL but I'm going to test a little more/look into adding tests - will update again when I'm confident to push.

emilymye:
Changed some of the login syntax/comments after the big command change and tested some more. I think it's ok to merge this now.

emilymye:
@jefferai updated!

PR #34, opened 2018-04-10
Make region name or self-link acceptable for bound_region

### LINKED ISSUE #33: bound_region is fully qualified URL, bound_zone is not
I genuinely don't know if this is working as intended or not.

I expected `bound_region` to support something like `europe-west4`. But it appears that the region actually has to be a fully qualified URL like: `https://www.googleapis.com/compute/v1/projects/my-project/regions/europe-west4`. Problem one I ran into is that isn't documented anywhere I found obvious... 

So that would be fine, but then I realized `bound_zone` _does accept_ `europe-west4-2` and _does not_ use fully qualified URLs. So this is a bit of a weird UX because two very similar bindings require two different formats for values and this isn't made clear.

I don't know if this is correct or not, but at best its a weird UX that we should probably document. Thanks!

emilymye:
Fix #33.

### DIFF
--- a/plugin/path_login.go
+++ b/plugin/path_login.go
@@ -504,27 +504,16 @@ func (b *GcpAuthBackend) authorizeGCEInstance(ctx context.Context, instance *com
 
 	// Verify that instance is in zone or region if given.
 	if len(role.BoundZone) > 0 {
-		var zone string
-		idx := strings.LastIndex(instance.Zone, "zones/")
-		if idx > 0 {
-			// Parse zone name from full zone self-link URL.
-			idx += len("zones/")
-			zone = instance.Zone[idx:len(instance.Zone)]
-		} else {
-			// Expect full zone name to be set as instance zone.
-			zone = instance.Zone
-		}
-
-		if zone != role.BoundZone {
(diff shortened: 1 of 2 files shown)


emilymye:
Added docs and decided we should probably also allow self-link for zone.

emilymye:
@jefferai @briankassouf PTAL again when you get the chance

plugin/path_login.go
```diff
@@ -584,6 +573,20 @@ func (b *GcpAuthBackend) authorizeGCEInstance(ctx context.Context, instance *com
 	return nil
 }
 
+func compareResourceNameOrSelfLink(expected, actual, collectionId string) bool {
+	sep := fmt.Sprintf("%s/", collectionId)
+	if strings.Contains(expected, sep) {
+		return expected == actual
+	}
+
+	expTkns := strings.SplitAfter(expected, sep)
```
kalafut: Are these two lines necessary? `expected` won't contain `sep` at this point, so I think `expectedName` is always the same as `expected`?

emilymye: fair

jefferai:
Looks good... 👍 to merge?

emilymye:
Please!

PR #40, opened 2018-07-03
Plumb through the logger

sethvargo:
This will display logs in the vault log output now 😄 

/cc @emilymye

### DIFF
--- a/plugin/backend.go
+++ b/plugin/backend.go
@@ -8,6 +8,7 @@ import (
 	"sync"
 
 	"github.com/hashicorp/go-gcp-common/gcputil"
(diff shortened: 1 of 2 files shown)


jefferai:
Is there a reason not to just use `b.Logger()`? As it's a new field nothing appears to actually be using `b.logger` right now and it should already be set up for you via that function.

sethvargo:
Is that new? I didn't think there was a way to access the logger without passing it through to the config struct?

jefferai:
It has to be in the config struct, but you're also passing it into Backend() and storing it locally (but then not actually accessing it). If it's in the config struct and goes into framework.Setup() then it's accessible via Logger().

Issue #27, opened 2018-01-16
Feature request: Support for multiple bound_zone, bound_region, bound_instance_group


ljagiello:
Hi,

Vault GCP plugin supports only single value for bound_zone, bound_region, bound_instance_group. Because of that limitation I cannot use a single role and I need to create duplicate roles for every combination of bound_zone, bound_region and bound_instance_group. It would be really great If I could still use bound_zone, bound_region, bound_instance_group but with multiple values.

Current state:
```
vault write auth/gcp/role/vault-us-west-1b type='gce' project_id="${GCE_PROJECT}" bound_zone="us-west1-b" bound_instance_group="vault-instance-group-manager-us-west1-b" policies="default,vault" max_ttl="${TTL}" ttl="${TTL}" period="${TTL}"
vault write auth/gcp/role/vault-us-west-1a type='gce' project_id="${GCE_PROJECT}" bound_zone="us-west1-a" bound_instance_group="vault-instance-group-manager-us-west1-a" policies="default,vault" max_ttl="${TTL}" ttl="${TTL}" period="${TTL}"
[…]
```
Desired state:
```
vault write auth/gcp/role/vault type='gce' project_id="${GCE_PROJECT}" bound_zone="us-west1-b,us-west1-a" bound_instance_group="vault-instance-group-manager-us-west1-b,vault-instance-group-manager-us-west1-a" policies="default,vault" max_ttl="${TTL}" ttl="${TTL}" period="${TTL}"
[…]
```

It will reduce complexity for any configuration in multiple regions/zones.

emilymye:
@ljagiello hi lukasz! we're been a little busy working on some other work for Vault-GCP integrations but I just wanted to ping to let you know I've seen this. Sorry for the delay!

ljagiello:
Thanks for info @emilymye !

ljagiello:
During the weekend I was working on some stuff connected with GCE and realize if `bound_instance_group` allow wildcards it would make any migration super easy.

Example:
```
vault write auth/gcp/role/worker-us-west-1b type='gce' project_id="${GCE_PROJECT}" bound_zone="us-west1-b" bound_instance_group="worker-instance-group-manager-us-west1-b" policies="default,worker" max_ttl="${TTL}" ttl="${TTL}" period="${TTL}"
```
[...]

PR #41, opened 2018-07-09
Support multiple binds and consistency updates

sethvargo:
This PR does a few things, primarily because I had to re-arrange some things to fix GH-27.

1. **Parallelize tests** - the tests now run in parallel. On the plane last night, they ran in ~8s. Today they are running at ~2s on my home Internet. That's a 5000% decrease in test time 😄! I had to add a few helpers to make sure tests wouldn't step on each other.

1. **Support multiple bindings for region, zone, and instance group** - this fixes GH-27 and provides users the ability to bind to multiple regions, zones, and/or instance groups instead of a single one. If a GCE instance exists in any of the given values, it will authenticate successfully. 

    Given the existing structure of the backend, it was very difficult to test this new logic in an independent way. I built an "authenticator" interface that makes the GCP API calls. For testing, those API calls can be stubbed with arbitrary values to test exclusively the logic. There's now a test for every edge case using these stubbed values that run very quickly to test our logic.

    The old schema is preserved and automatically upgraded the first time a role is accessed. This is inline with the recommendation from the Vault team on how to handle such upgrades. 

1. **Warning passthrough** - theres now a way to pass warnings up from the helper methods, which is helpful during decoding, etc.

1. **Validate input fields** - this is something Vault is doing sporadically, but I added it here everywhere. Nothing is more frustrating than when you accidentally mistype a field (like `vault write auth/gcp/config cradentials=...`) and Vault happily accepts that. Fields that don't match the schema will now inform the user of their error, hopefully saving valuable debugging time.

1. **Reduce the number of API calls** - there are some optimizations to reduce the number of API calls, including server-side filter and client-side string parsing. For example, given a zone, we can **always** calculate the region from the zone using string parsing; we don't have to incur an API. The zone/regions spec specifically documents this format, so we are okay to use it. On average, this saves two API calls per (GCE) authentication.

🚧 This will require a docs update in the main Vault repo. I can take on that work once we are close to merging this.

### DIFF
--- a/plugin/authorizer_client_gcp.go
+++ b/plugin/authorizer_client_gcp.go
@@ -1,0 +1,81 @@
+package gcpauth
+
+import (
+	"context"
+	"fmt"
+	"log"
+
+	"github.com/hashicorp/vault/helper/strutil"
+	compute "google.golang.org/api/compute/v1"
+	iam "google.golang.org/api/iam/v1"
+)
+
+var _ client = (*gcpClient)(nil)
+
+// gcpClient implements client and communicates with the GCP API. It is
+// abstracted as an interface for stubbing during testing. See stubbedClient for
+// more details.
+type gcpClient struct {
+	computeSvc *compute.Service
+	iamSvc     *iam.Service
+}
+
+func (c *gcpClient) InstanceGroups(ctx context.Context, project string, boundInstanceGroups []string) (map[string][]string, error) {
+	// map of zone names to a slice of instance group names in that zone.
+	igz := make(map[string][]string)
+
+	if err := c.computeSvc.InstanceGroups.
+		AggregatedList(project).
+		Fields("items/*/instanceGroups/name").
+		Pages(ctx, func(l *compute.InstanceGroupAggregatedList) error {
+			for k, v := range l.Items {
+				zone, err := zoneFromSelfLink(k)
+				if err != nil {
+					return err
+				}
+
+				for _, g := range v.InstanceGroups {
+					if strutil.StrListContains(boundInstanceGroups, g.Name) {
+						igz[zone] = append(igz[zone], g.Name)
+					}
+				}
+			}
+			return nil
+		}); err != nil {
+		return nil, err
+	}
+
+	return igz, nil
+}
+
+func (c *gcpClient) InstanceGroupContainsInstance(ctx context.Context, project, zone, group, instanceSelfLink string) (bool, error) {
+	var req compute.InstanceGroupsListInstancesRequest
+	resp, err := c.computeSvc.InstanceGroups.
+		ListInstances(project, zone, group, &req).
+		Filter(fmt.Sprintf("instance eq %s", instanceSelfLink)).
+		Context(ctx).
+		Do()
+	if err != nil {
+		return false, err
+	}
+
+	log.Printf("%#v", resp.Items)
+
+	if resp != nil && len(resp.Items) > 0 {
+		return true, nil
+	}
+	return false, nil
+}
+
+func (c *gcpClient) ServiceAccount(ctx context.Context, name string) (string, string, error) {
+	account, err := c.iamSvc.Projects.ServiceAccounts.
+		Get(name).
+		Fields("uniqueId", "email").
+		Context(ctx).
+		Do()
+	if err != nil {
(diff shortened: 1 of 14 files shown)


plugin/path_role.go
```diff
@@ -7,7 +7,9 @@ import (
 	"strings"
 	"time"
 
+	"github.com/fatih/structs"
```
emilymye: We're trying to avoid using third-party libraries that belong to one person like this as much as possible. Also, I think @jefferai mentioned Vault was moving specifically away from using structs for converting the response and just writing it out instead. 

Otherwise, this requires updates to deps

plugin/path_role.go
```diff
@@ -261,38 +282,22 @@ func (b *GcpAuthBackend) pathRoleRead(ctx context.Context, req *logical.Request,
...
-		}
-	}
+	role.Period /= time.Second
+	role.TTL /= time.Second
+	role.MaxTTL /= time.Second
+	role.MaxJwtExp /= time.Second
 
 	return &logical.Response{
-		Data: roleMap,
+		Data: structs.New(role).Map(),
```
emilymye: per previous comment, don't use structs here

plugin/authorizer_client_gcp.go
```diff
@@ -0,0 +1,81 @@
+package gcpauth
+
+import (
+	"context"
+	"fmt"
+	"log"
+
+	"github.com/hashicorp/vault/helper/strutil"
+	compute "google.golang.org/api/compute/v1"
```
emilymye: nit/note - I don't think you actually need to alias these imports?

plugin/authorizer_client_gcp.go
```diff
@@ -0,0 +1,81 @@
...
+	iamSvc     *iam.Service
+}
+
+func (c *gcpClient) InstanceGroups(ctx context.Context, project string, boundInstanceGroups []string) (map[string][]string, error) {
+	// map of zone names to a slice of instance group names in that zone.
+	igz := make(map[string][]string)
+
+	if err := c.computeSvc.InstanceGroups.
+		AggregatedList(project).
+		Fields("items/*/instanceGroups/name").
```
emilymye: We should probably use [filters](https://cloud.google.com/compute/docs/reference/rest/v1/instances/aggregatedList#filter) for name and zones  instead of having the pagination function?

emilymye:
>Support pagination for long GCP API responses - could be viewed as a bug, but when the list of instance groups was really long, we weren't paginating. This is fixed.

 A little confused - were we listing instance groups?

plugin/authorizer_client_gcp.go
sethvargo: It's not possible due to a bug in the API: https://stackoverflow.com/questions/50452838/google-cloud-compute-instances-aggregatedlist-with-filter-fails

sethvargo:
> A little confused - were we listing instance groups?

Apparently not. My mistake. I've updated the original issue to remove that.

plugin/authorizer_client_gcp.go
sethvargo: We'd also still need the pagination function, even if we used filters. Filters are applied _after_ the response is generated, so you'd still need to paginate, potentially though a bunch of empty pages. For example, if there are 400 results, and you filter, the first 20 pages might be empty, since they _were_ filled prior to filtering.

But yea - you can test trying to filter in the docs online. It's broken. There's an internal ticket about it.

sethvargo:
@emilymye updated per your other feedback

emilymye:
@sethvargo thanks for doing this! It mostly LGTM but I'm going to take a little longer to review this since we actually don't have tests for the GCE auth right now (something I'm actually trying to get set up seperately now - turns out the fact that GCE metadata must be accessed via an instance makes testing hard and slow, who knew)

sethvargo:
@emilymye yea, unfortunately I'm aware. If it helps, I've tested it manually on a few different GCE instances, but it would be great to have an automated way to test stuff

plugin/authorizer_client_gcp.go
emilymye: wow what a BIG OLE BUNCH OF FUN :(
lgtm then

PR #45, opened 2018-08-07
Use clean background ctx

emilymye:


briankassouf:
Is there any way for the individual request scoped gcp calls to use the request context? It would be nice if we could cancel them if for some reason the network call is blocked.

emilymye:
Most of the calls are in hashicorp/go-gcp-common so I'd be doing it there

PR #52, opened 2018-10-04
Rename gcpRole struct field from "RoleType" to "Type", to align with the input

broamski:
This change renames the gcpRole struct field "RoleType" to "Type", which aligns the returned config schema to the input schema and eliminates some language redundancy

e.g. `"type": "iam"` is used to create/configure the role, meanwhile reading a role's configuration returns `"role_type": "iam`

The schema mismatch is problematic when using a vault configuration management tool as the returned result always differs from the desired state.

### DIFF
--- a/plugin/path_role.go
+++ b/plugin/path_role.go
@@ -289,7 +289,7 @@ func (b *GcpAuthBackend) pathRoleRead(ctx context.Context, req *logical.Request,
 	resp := make(map[string]interface{})
 
 	if role.RoleType != "" {
-		resp["role_type"] = role.RoleType
+		resp["role"] = role.RoleType
 	}
 	if role.ProjectId != "" {
 		resp["project_id"] = role.ProjectId


plugin/path_role.go
```diff
@@ -288,8 +288,8 @@ func (b *GcpAuthBackend) pathRoleRead(ctx context.Context, req *logical.Request,
 
 	resp := make(map[string]interface{})
 
-	if role.RoleType != "" {
-		resp["role_type"] = role.RoleType
+	if role.Type != "" {
+		resp["type"] = role.Type
```
emilymye: A quicker fix is to undo all and just change this string value "role_type" to "type". I'd prefer this, just because there are fewer places where you could possibly miss a reference to role_type/RoleType. If you can also change https://github.com/hashicorp/vault-plugin-auth-gcp/blob/master/plugin/path_role_test.go and run the test (`make test TEST=./plugin`) that would be helpful! Thank you.

emilymye: Also, possibly more importantly, this avoids errors when reading role structs/JSON stored in Vault storage from previous versions of the plugin.

emilymye:
@jefferai @briankassouf is it imortant to keep the `role_type` output parameter as well (or instead?). Can we consider this not an important/breaking change?

plugin/path_role.go
broamski: Ah, good call. I originally did this but talked myself out of it. I'll update the PR!

broamski:
If there is a concern  about backwards compatibility, I can change this so that we can return both `role_type` and `type`.

broamski:
Hi! Checking in - is this good as-is?

PR #55, opened 2018-11-07
Group aliases, project_id --> optional bound_projects, no caching GCP API clients

emilymye:
- Add ability to enable group aliases (opt-in because additional permissions are required)
- Deprecate required field project_id, add optional field bound_projects; if included, authenticating entity must belong to bound_projects
- No caching GCP API clients: Removes token staleness issue and in general I don't think it wasn't necessary
- Added local_dev script for ease of testing.

### DIFF
--- a/plugin/backend.go
+++ b/plugin/backend.go
@@ -5,16 +5,18 @@ import (
 	"fmt"
 	"github.com/hashicorp/errwrap"
 	"github.com/hashicorp/go-cleanhttp"
-	"github.com/hashicorp/go-gcp-common/gcputil"
+	"github.com/hashicorp/vault/helper/useragent"
 	"github.com/hashicorp/vault/logical"
 	"github.com/hashicorp/vault/logical/framework"
 	"golang.org/x/oauth2"
-	"golang.org/x/oauth2/jwt"
+	"golang.org/x/oauth2/google"
 	"google.golang.org/api/cloudresourcemanager/v1"
 	"google.golang.org/api/compute/v1"
 	"google.golang.org/api/iam/v1"
 	"net/http"
 )
+
+const defaultCloudScope = "https://www.googleapis.com/auth/cloud-platform"
 
 type GcpAuthBackend struct {
 	*framework.Backend
@@ -34,11 +36,7 @@ func Factory(ctx context.Context, conf *logical.BackendConfig) (logical.Backend,
 
 func Backend() *GcpAuthBackend {
 	b := &GcpAuthBackend{
-		oauthScopes: []string{
-			iam.CloudPlatformScope,
-			compute.ComputeReadonlyScope,
-			cloudresourcemanager.CloudPlatformScope,
-		},
+		oauthScopes: []string{defaultCloudScope},
 	}
 
 	b.Backend = &framework.Backend{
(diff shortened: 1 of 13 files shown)


plugin/backend.go
```diff
@@ -45,13 +37,13 @@ func Backend() *GcpAuthBackend {
 		oauthScopes: []string{
 			iam.CloudPlatformScope,
 			compute.ComputeReadonlyScope,
+			cloudresourcemanager.CloudPlatformScope,
```
sethvargo: Let's make sure we document this change in scope in the website docs and changelog

plugin/backend.go
```diff
@@ -72,106 +64,37 @@ func Backend() *GcpAuthBackend {
...
-}
-
-// Initialize attempts to create GCP clients from stored config.
-// It does not attempt to claim the client lock.
-func (b *GcpAuthBackend) initClients(ctx context.Context, s logical.Storage) (err error) {
+func (b *GcpAuthBackend) httpClient(ctx context.Context, s logical.Storage) (*http.Client, error) {
 	config, err := b.config(ctx, s)
 	if err != nil {
-		return err
+		return nil, fmt.Errorf("credentials were not configured and fallback to application default credentials failed: %v", err)
```
sethvargo: 1. Line wrap
1. This lacks a call to action - what should the user do? How can they fix this? Where can they look for more information?
1. Use `errwrap`

plugin/backend.go
```diff
@@ -72,106 +64,37 @@ func Backend() *GcpAuthBackend {
...
-		}
-	}
-
-	return b.gceClient, nil
-}
-
-// Initialize attempts to create GCP clients from stored config.
-// It does not attempt to claim the client lock.
-func (b *GcpAuthBackend) initClients(ctx context.Context, s logical.Storage) (err error) {
+func (b *GcpAuthBackend) httpClient(ctx context.Context, s logical.Storage) (*http.Client, error) {
```
sethvargo: I think you might be able to [simplify this a bit](https://github.com/hashicorp/vault-plugin-secrets-gcpkms/blob/a7d4151e40a5c1649d41440cadd1fe654df7e61c/backend.go#L154-L169)

plugin/backend.go
```diff
@@ -72,106 +64,37 @@ func Backend() *GcpAuthBackend {
...
+		client := conf.Client(ctx)
+		return client, nil
 		httpClient, err = gcputil.GetHttpClient(config.Credentials, b.oauthScopes...)
 		if err != nil {
-			return err
+			return nil, errwrap.Wrapf("could not create HTTP client for given config credentials: {{err}}", err)
 		}
 	}
-
-	userAgentStr := fmt.Sprintf("(%s %s) Vault/%s", runtime.GOOS, runtime.GOARCH, version.GetVersion().FullVersionNumber(true))
```
sethvargo: Make sure we don't lose this - we should set the user agent. There's also a helper now (`useragent.String()`).

plugin/cli.go
```diff
@@ -163,8 +163,8 @@ Configuration:
 	permissions on this service account. 
   
   project=<string>                                
-	Project the service account belongs to. Defaults to credentials 
-	"project_id" if "credentials" specified and this value is not.  
+	Project for the service account who will be authenticating to Vault.
+    Defaults to credentials "project_id" if credentials are specified.
```
sethvargo: "Defaults to the credential's "project_id" if credentials are specified."

plugin/path_login.go
```diff
@@ -472,14 +520,52 @@ func (b *GcpAuthBackend) pathGceLogin(ctx context.Context, req *logical.Request,
...
+		aliases, err := b.groupAliases(httpC, ctx, metadata.ProjectId)
+		if err != nil {
+			return nil, err
+		}
+		resp.Auth.GroupAliases = aliases
+	}
 	return resp, nil
 }
 
+func (b *GcpAuthBackend) groupAliases(httpClient *http.Client, ctx context.Context, projectId string) ([]*logical.Alias, error) {
```
sethvargo: Please document this and describe the expected response based on sample data

plugin/path_login_test.go
```diff
@@ -34,12 +34,14 @@ func TestLoginIam(t *testing.T) {
 	})
 
 	roleName := "testrole"
+	projects := []string{creds.ProjectId, "someproject"}
 	testRoleCreate(t, b, reqStorage, map[string]interface{}{
 		"name":                   roleName,
 		"type":                   "iam",
 		"policies":               "dev, prod",
-		"project_id":             creds.ProjectId,
```
sethvargo: Let's keep/add a test with `project_id`. We don't want to break existing setups/automation that use `project_id`

plugin/path_role.go
```diff
@@ -960,3 +976,27 @@ func checkInvalidRoleTypeArgs(data *framework.FieldData, invalidSchema map[strin
...
+
+// deprecatedFieldSchema contains the deprecated role attributes
+var deprecatedFieldSchema = map[string]*framework.FieldSchema{
+	"service_accounts": {
+		Type:        framework.TypeCommaStringSlice,
+		Description: `Deprecated, use bound_service_accounts instead.`,
+	},
+	"project_id": {
+		Type:        framework.TypeString,
+		Description: `The id of the project that authorized instances must belong to for this role.`,
```
sethvargo: Prefix with `Deprecated:`

plugin/path_role.go
```diff
@@ -960,3 +976,27 @@ func checkInvalidRoleTypeArgs(data *framework.FieldData, invalidSchema map[strin
 	}
 	return nil
 }
+
+// deprecatedFieldSchema contains the deprecated role attributes
+var deprecatedFieldSchema = map[string]*framework.FieldSchema{
+	"service_accounts": {
+		Type:        framework.TypeCommaStringSlice,
+		Description: `Deprecated, use bound_service_accounts instead.`,
```
sethvargo: `Deprecated:` (not comma)

sethvargo: `\"bound_service_accounts\"`

plugin/path_role.go
sethvargo: Rephrase to suggest replacement

plugin/backend.go
emilymye: I'm also not really sure we should do this at all - iam.CloudPlatformScope and cloudresourcemanager.CloudPlatformScope are both `https://www.googleapis.com/auth/cloud-platform` which is everything? I don't really know if any of the GCP go libraries have a standard "cloud platform scope" constant that I can just use.

sethvargo: `cloud-platform` is the everything scope

sethvargo: I usually just make my own constant: https://github.com/hashicorp/vault-plugin-secrets-gcpkms/blob/a7d4151e40a5c1649d41440cadd1fe654df7e61c/config.go#L11

plugin/path_login_test.go
emilymye: Since there is a test for this in path_role_test.go, I am going to just not keep the deprecated value in the login tests.

PR #63, opened 2019-03-14
Cache clients to improve performance

sethvargo:
Our current implementation creates 3 new clients on each login request. For a large set of concurrent requests, this can exhaust FDs and dramatically slow down performance.

This PR implements client caching to cache each client instantiation for 30min (half of the 60min jwt expiration time). In local testing, this resulted in a 612x speedup when authenticating via IAM at 300 req/min for 5 minutes.

### DIFF
--- a/plugin/backend.go
+++ b/plugin/backend.go
@@ -2,9 +2,12 @@ package gcpauth
 
 import (
 	"context"
-	"fmt"
+	"net/http"
+	"time"
+
 	"github.com/hashicorp/errwrap"
 	"github.com/hashicorp/go-cleanhttp"
+	"github.com/hashicorp/vault-plugin-auth-gcp/plugin/cache"
 	"github.com/hashicorp/vault/helper/useragent"
 	"github.com/hashicorp/vault/logical"
 	"github.com/hashicorp/vault/logical/framework"
@@ -13,16 +16,20 @@ import (
 	"google.golang.org/api/cloudresourcemanager/v1"
 	"google.golang.org/api/compute/v1"
 	"google.golang.org/api/iam/v1"
-	"net/http"
 )
 
-const defaultCloudScope = "https://www.googleapis.com/auth/cloud-platform"
+var (
+	// cacheTime is the duration for which to cache clients and credentials. This
+	// must be less than 60 minutes.
+	cacheTime = 30 * time.Minute
+)
 
 type GcpAuthBackend struct {
 	*framework.Backend
 
-	// OAuth scopes for generating HTTP and GCP service clients.
-	oauthScopes []string
+	// cache is the internal client/object cache. Callers should never access the
+	// cache directly.
+	cache *cache.Cache
 }
 
 // Factory returns a new backend as logical.Backend.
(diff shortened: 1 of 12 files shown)


sethvargo:
@jefferai @briankassouf @kalafut can one of you please review and merge? Thanks!

plugin/cache/cache.go
```diff
@@ -0,0 +1,87 @@
...
+// Fetch retrieves an item from the cache. If the item exists in the cache and
+// is within its lifetime, it is returned. If the item does not exist, or if the
+// item exists but has exceeded its lifetime, the function f is invoked and the
+// result is updated in the cache.
+func (c *Cache) Fetch(name string, t time.Duration, f Func) (interface{}, error) {
+	// Attempt to read from the cache, returning the cached value if it's still
+	// valid.
+	c.lock.RLock()
+	e, ok := c.data[name]
+	if ok && e.result != nil && time.Now().UTC().Sub(e.created) < e.lifetime {
```
kalafut: The `UTC()` conversion isn't needed in these calculations (or the time storage).

sethvargo:
@emilymye @kalafut removed the calls to `UTC`. Let me know if there's anything else!

plugin/backend.go
```diff
@@ -58,94 +65,167 @@ func Backend() *GcpAuthBackend {
...
+		Invalidate: b.invalidate,
 	}
 	return b
 }
 
-func (b *GcpAuthBackend) httpClient(ctx context.Context, s logical.Storage) (*http.Client, error) {
-	config, err := b.config(ctx, s)
+// IAMClient returns a new IAM client. The client is cached.
+func (b *GcpAuthBackend) IAMClient(s logical.Storage) (*iam.Service, error) {
+	httpClient, err := b.httpClient(s)
```
briankassouf: Could this be moved inside the fetch closure? (and in the other functions)

plugin/backend.go
```diff
@@ -58,94 +65,167 @@ func Backend() *GcpAuthBackend {
...
 	if err != nil {
-		return nil, fmt.Errorf(clientErrorTemplate, "IAM", err)
+		return nil, errwrap.Wrapf("failed to create oauth2 http client: {{err}}", err)
 	}
-	iamClient.UserAgent = useragent.String()
 
-	gceClient, err := compute.New(httpC)
+	client, err := b.cache.Fetch("HTTPClient", cacheTime, func() (interface{}, error) {
+		b.Logger().Debug("creating oauth2 http client")
+		ctx := context.WithValue(context.Background(), oauth2.HTTPClient, cleanhttp.DefaultClient())
```
kalafut: Since we're redoing this, perhaps we should incorporate`context.WithCancel()`, save the `CancelFunc` on backend object, and implement `Clean()` to invoke it if set.

plugin/backend.go
sethvargo: @briankassouf it cannot because the HTTP client itself is also inside a fetch closure, which would create a deadlock. The first invocation of `b.httpClient` will create the cache.

plugin/backend.go
sethvargo: I had that originally before I built the `cache` package, but I think it'd be really complicated with the race conditions. `Clean()` would have to acquire a lock on the HTTP client key before calling cancel, and I'm not sure it's worth the complexity given how quickly most of these requests will complete?

kalafut: It can be left as is.

jefferai:
@sethvargo should this be made in the secrets engine too?

sethvargo:
@jefferai https://github.com/hashicorp/vault-plugin-secrets-gcp/pull/33

PR #67, opened 2019-07-01
Add tolerance to expiration checks

### LINKED ISSUE #54: Sporadic login failures - role requires that JWTs must expire within 900 seconds
I occasionally get an error message when using the Vault CLI to login with the GCP auth method.  If I run the same command over and over, it fails some of the time and succeeds other times.
```no-highlight
> vault login -method=gcp role=provisioner project=<PROJECT> service_account=<EMAIL>
Error authenticating: Error making API request.

URL: PUT https://<OBFUSCATED>/v1/auth/gcp/login
Code: 400. Errors:

* role requires that JWTs must expire within 900 seconds
```

If I set my local computer's clock ahead just 2 seconds, it seems to fail every time.  On the other hand, if I add `jwt_exp=10` to my command line, then I'm consistently able to log in without receiving that error message.

I suspect this is because the default `max_jwt_exp` on GCP auth roles is 15 minutes, and the CLI also by default requests JWTs with an expiration 15 minutes from now.  It seems like slight clock drift between the client and Vault server could cause these inconsistent results when the defaults are used.

There's an obvious workaround of increasing the `max_jwt_exp` on every role or always passing a shorter `jwt_exp` to the CLI, but it would be nice if the defaults worked consistently.  I wonder if we should consider adding a small tolerance when validating the expiration (https://github.com/hashicorp/vault-plugin-auth-gcp/blob/a00186a9031fc8db194e9b6002eed76db0ded90c/plugin/path_login.go#L303) or changing the Vault CLI to use a default `jwt_exp` slightly shorter than `15` (https://github.com/hashicorp/vault/blob/903ab7c4858085e4ef8c4cb05d071bd5f3bdbe10/command/agent/auth/gcp/gcp.go#L27)?

### Environment
I'm running Vault in GCP in a GKE cluster using the vault:0.11.4 image from DockerHub.  The client I'm authenticating from is a Windows laptop on a corporate network running the Windows x64 Vault client v0.11.4.

emilymye:
Fixes #54

### DIFF
--- a/plugin/path_login.go
+++ b/plugin/path_login.go
@@ -22,6 +22,7 @@ import (
 
 const (
 	expectedJwtAudTemplate string = "vault/%s"
+	jwtExpToleranceSec = 60
 )
 
 func pathLogin(b *GcpAuthBackend) *framework.Path {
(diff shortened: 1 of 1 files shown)


jefferai:
I think this looks fine although 1 second is...not much. I believe Square's jwt library defaults to a minute or so.

emilymye:
oh ok! I'll can do more, just wasn't sure what a reasonable tolerance was

PR #74, opened 2019-09-27
Fix regional ig issue on InstanceGroup.AggregatedList

### LINKED ISSUE #70: `failed to list instance groups for project \"<project>": failed to extract zone from self-link \"regions/europe-west2\"}"`
Hi :wave: 

Thanks for the awesome project! 

We are having periodic errors from our vault clients attempting to login via GCP:
```
PUT http://<IP>:8200/v1/auth/gcp/login\nCode: 400. Errors:\n\n* failed to list instance groups for project \"<project>\": failed to extract zone from self-link \"regions/europe-west2\"}"
```
It is flaky. Sometimes this call works, sometimes not.

We are using:
* GCE VMs
* `vault 1.1.0`  which is built with this `auth-gcp` plugin from commit `7d4c2101e7d0b61ec9fb0dc3c75d79920c6369c5`

Plugin code did not change from that point of time.

If we are not wrong this endpoint is used (?) by `compute/v1` plugin is using:
`https://www.googleapis.com/compute/v1/projects/<project>/aggregated/instanceGroups` 

And it indeed returns both keys `regions/xxx` as well as `zones/xxx` in response.

Any clue what might be happening? (:

emilymye:
Fixes #70

### DIFF
--- a/plugin/authorizer_client_gcp.go
+++ b/plugin/authorizer_client_gcp.go
@@ -3,9 +3,9 @@ package gcpauth
 import (
 	"context"
 	"fmt"
-	"log"
 	"strings"
 
+	log "github.com/hashicorp/go-hclog"
 	"github.com/hashicorp/vault/sdk/helper/strutil"
 	"google.golang.org/api/compute/v1"
 	"google.golang.org/api/iam/v1"
@@ -17,6 +17,7 @@ var _ client = (*gcpClient)(nil)
 // abstracted as an interface for stubbing during testing. See stubbedClient for
 // more details.
 type gcpClient struct {
+	logger     log.Logger
 	computeSvc *compute.Service
 	iamSvc     *iam.Service
 }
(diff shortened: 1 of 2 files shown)


emilymye:
@improbable-ludwik

plugin/authorizer_client_gcp.go
```diff
@@ -30,7 +30,9 @@ func (c *gcpClient) InstanceGroups(ctx context.Context, project string, boundIns
 			for k, v := range l.Items {
 				zone, err := zoneFromSelfLink(k)
 				if err != nil {
-					return err
```
sethvargo: I don't like swallowing this error. Can we check for a substring of `/regions` in `k` before calling `zoneFromSelfLink` instead?

```go
if strings.Contains(k, "/regions") {
  log.Printf("got back a region, dunno why")
  continue
}

zone, err := zoneFromSelfLink(k)
if err != nil {
  return err
}
```

plugin/authorizer_client_gcp.go
```diff
@@ -28,6 +30,13 @@ func (c *gcpClient) InstanceGroups(ctx context.Context, project string, boundIns
 		Fields("items/*/instanceGroups/name").
 		Pages(ctx, func(l *compute.InstanceGroupAggregatedList) error {
 			for k, v := range l.Items {
+				// Some groups returned are regional
+				// TODO(emilymye, #73): Support regions?
+				if strings.Contains(k, "/regions/") {
+					log.Printf("[WARN] ignoring key %q with region in instance group aggregated list", k)
```
kalafut: Preferable to add the backend's logger to `gcpClient` and log to that instead.

kalafut:
@emilymye LGTM except for the merge conflicts that crept in.

kalafut:
Thanks!

PR #77, opened 2019-12-13
Get service account via public URL, fix GCE inferrence bug

emilymye:
- Use the public service account key URL
- Fix JWT exp check to not check GCE tokens (since they don't allow for custom exp)

Dependent on re-mod after https://github.com/hashicorp/go-gcp-common/pull/3


plugin/path_login.go
```diff
@@ -163,12 +167,12 @@ func (b *GcpAuthBackend) parseAndValidateJwt(ctx context.Context, req *logical.R
...
 	jwtVal, err := jwt.ParseSigned(signedJwt.(string))
 	if err != nil {
-		return nil, err
+		return nil, errwrap.Wrapf("unable to parse signed JWT: {{err}}", err)
 	}
 
 	key, err := b.getSigningKey(ctx, jwtVal, signedJwt.(string), loginInfo.Role, req.Storage)
 	if err != nil {
-		return nil, err
+		return nil, errwrap.Wrapf("unable to get public key for signed JWT: %v", err)
```
kalafut:
```suggestion
		return nil, errwrap.Wrapf("unable to get public key for signed JWT: {{err}}", err)
```

plugin/path_login.go
```diff
@@ -313,8 +296,10 @@ func (b *GcpAuthBackend) pathIamLogin(ctx context.Context, req *logical.Request,
 	}
 
 	// TODO(emilymye): move to general JWT validation once custom expiry is supported for other JWT types.
-	if loginInfo.JWTClaims.Expiry.Time().After(time.Now().Add(role.MaxJwtExp)) {
-		return logical.ErrorResponse(fmt.Sprintf("role requires that JWTs must expire within %d seconds", int(role.MaxJwtExp/time.Second))), nil
+	if loginInfo.GceMetadata != nil {
+		b.Logger().Info("GCE Metadata found in JWT, skipping custom expiry check")
+	} else if loginInfo.JWTClaims.Expiry.Time().After(time.Now().Add(role.MaxJwtExp)) {
+		return logical.ErrorResponse(fmt.Sprintf("role requires that service account JWTs expire within %d seconds", int(role.MaxJwtExp/time.Second))), nil
```
kalafut:
```suggestion
		return logical.ErrorResponse("role requires that service account JWTs expire within %d seconds", int(role.MaxJwtExp/time.Second)), nil
```

scripts/dev.sh
```diff
@@ -0,0 +1,52 @@
...
+  -dev \
+  -dev-plugin-init \
+  -dev-plugin-dir "${SCRATCH}/plugins" \
+  -dev-root-token-id "root" \
+  -log-level "debug" \
+  &
+sleep 2
+VAULT_PID=$!
+
+echo "    Mouting plugin"
```
kalafut:
```suggestion
echo "    Mounting plugin"
```
🙂

emilymye: ; __ ;

PR #84, opened 2020-03-11
Add support for regional instance groups

### LINKED ISSUE #73: Support regional instance groups for bound_instance_group
Would be an enhancement, currently we only support normal (zonal) IG

dhduvall:
Fixes #73

### DIFF
--- a/plugin/authorizer_client_gcp.go
+++ b/plugin/authorizer_client_gcp.go
@@ -3,7 +3,6 @@ package gcpauth
 import (
 	"context"
 	"fmt"
-	"strings"
 
 	log "github.com/hashicorp/go-hclog"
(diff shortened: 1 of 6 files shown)


willdeuschle:
Can this be merged? GCE auth won't work without it.

willdeuschle:
To provide a data point in support of this PR - I forked the plugin, applied this patch, built vault with my forked plugin, and as expected I can now use GCE auth to login via vault.

robmonte (approved):
Everything looks good to me! Ran through the flow without any issues. Interestingly enough, I don't even think the docs need an update to accommodate this feature as the current wording already mentions bound instance groups via regions.

PR #88, opened 2020-04-20
Make identities less unique to improve performance

dustin-decker:
Signed-off-by: Dustin Decker <dustindecker@protonmail.com>

# Overview
This change removes instance display names and metadata from GCP auth login responses. This increase performance at the expense of tracability.

When an alias is unique, Vault has to read a StoragePacker bucket, decompress it, unmarshal the protobuf, modify it, marshal it back to a protobuf, compress it, and write the bucket back. During any StoragePacker operation, a lock is put on that key which means that only one operation on the bucket can happen at any time. These operations are very expensive and get more expensive as the entity count grows.

Our entity count from the GCP auth backend is around 5 million currently, so each login results in processing ~22,300 entities in one of the StoragePacker buckets with an exclusive storage lock on that bucket. This results Vault failing if more than ~10 GCP authentications are being performed at a given time. We need to be able to do thousands per minute.

https://github.com/hashicorp/vault/issues/8761

# Contributor Checklist
[My Docs PR Link](link)
[ ] Backwards compatible

### DIFF
--- a/plugin/path_login.go
+++ b/plugin/path_login.go
@@ -453,7 +453,7 @@ func (b *GcpAuthBackend) pathGceLogin(ctx context.Context, req *logical.Request,
 		return &logical.Response{
 			Auth: &logical.Auth{
 				Alias: &logical.Alias{
-					Name: fmt.Sprintf("gce-%s", strconv.FormatUint(instance.Id, 10)),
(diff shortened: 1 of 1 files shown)


dustin-decker:
Closing in favor of #89

PR #89, opened 2020-04-22
Add alias types for IAM and GCE logins

pcman312:
# Overview
Allows users to specify an alias type field for IAM and GCE logins which will then switch between the current behavior (a unique ID for IAM, and the instance ID for GCE) and a newly created `role_id` field. The `role_id` is a UUID generated when the role is created. All existing roles without a role_id will have one generated and saved when the role is read or written.

# Design of Change
Adds 3 fields to the role config:
- `role_id` (generated)
- `iam_alias`- `"unique_id"` (default) or `"role_id"`
- `gce_alias`- `"instance_id"` (default) or `"role_id"`

If the alias field is specified as `"role_id"`, the Vault alias will be the generated `role_id` rather than the current behavior of a unique ID (for IAM) or the instance ID (for GCE).

I've also utilized go's mocking library to mock the `logical.SystemView` and `logical.Storage` interfaces to improve testing coverage of some new tests. To generate the mocks, run `make mocks`.

The `gcpRole` struct and its associated functions have been pulled out into its own file: `gcp_role.go` since the `path_role.go` file was getting very large (1000+ lines).

- [x] Resolves https://github.com/hashicorp/vault/issues/8761
- [x] Docs PR: https://github.com/hashicorp/vault/pull/8822

# Test Output
```shell
$ make test
ok  	github.com/hashicorp/vault-plugin-auth-gcp/plugin	2.946s
```

# Contributor Checklist
- [x] Add relevant docs to upstream Vault repository, or sufficient reasoning why docs won’t be added yet
- [x] Add output for any tests not ran in CI to the PR description (eg, acceptance tests)
- [x] Backwards compatible


plugin/path_role.go
```diff
@@ -480,8 +502,8 @@ func (b *GcpAuthBackend) pathRoleEditGceLabels(ctx context.Context, req *logical
 	return b.storeRole(ctx, req.Storage, roleName, role, warnings)
 }
 
-// role reads a gcpRole from storage. This assumes the caller has already obtained the role lock.
-func (b *GcpAuthBackend) role(ctx context.Context, s logical.Storage, name string) (*gcpRole, error) {
+// retrieveRole from storage. This assumes the caller has already obtained the role lock.
+func (b *GcpAuthBackend) retrieveRole(ctx context.Context, s logical.Storage, name string) (*gcpRole, error) {
```
catsby, line 506: Can you share the rational for renaming this method?

plugin/path_role_test.go
```diff
@@ -451,6 +464,520 @@ func TestRole_InvalidRoleType(t *testing.T) {
...
+		getResp  *logical.StorageEntry
+		getErr   error
+		getTimes int
+
+		localMount            bool
+		localMountTimes       int
+		replicationState      consts.ReplicationState
+		replicationStateTimes int
+
+		putErr   error
```
catsby, line 481: It looks like `putErr` is `nil` in all of the test cases, am I reading that correctly?

plugin/path_role.go
pcman312: Generally I ascribe to the idea that function names should be read like verbs.

plugin/path_login.go
```diff
@@ -475,7 +487,7 @@ func (b *GcpAuthBackend) pathGceLogin(ctx context.Context, req *logical.Request,
 	auth := &logical.Auth{
 		InternalData: map[string]interface{}{},
 		Alias: &logical.Alias{
-			Name: fmt.Sprintf("gce-%s", strconv.FormatUint(instance.Id, 10)),
+			Name: alias,
```
tyrannosaurus-becks, line 490: One interesting thing here is that, the alias _could_ be the `role_id`. However, the alias `metadata` fields don't change no matter what they select. Some of the fields would have a high rate of change if `role_id` were selected (`instance_name`, for example). This means if they selected `role_id`, they could have a performance hit.

@kalafut is this something we should merely document in this PR? Or should this PR (or a follow-on one) include the ability to configure metadata fields using the helper package being introduced in https://github.com/hashicorp/vault/pull/8783?

plugin/path_role_test.go
pcman312: Good catch. Removed. Storage errors were originally in the table tests but I pulled it out into its own test and neglected to remove it here.

plugin/gcp_role.go
```diff
@@ -0,0 +1,429 @@
...
+
+	// Set role type
+	if rt, ok := data.GetOk("type"); ok {
+		roleType := rt.(string)
+		if role.RoleType != roleType && req.Operation == logical.UpdateOperation {
+			return warnings, fmt.Errorf("role type cannot be changed for an existing role")
+		}
+		role.RoleType = roleType
+	} else if req.Operation == logical.CreateOperation {
+		return warnings, fmt.Errorf(errEmptyRoleType)
```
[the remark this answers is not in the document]
pcman312: Done.

plugin/gcp_role.go
```diff
@@ -0,0 +1,429 @@
...
+		if e := tokenutil.UpgradeValue(data, "period", "token_period", &role.Period, &role.TokenPeriod); e != nil {
+			return nil, e
+		}
+	}
+
+	// Set role type
+	if rt, ok := data.GetOk("type"); ok {
+		roleType := rt.(string)
+		if role.RoleType != roleType && req.Operation == logical.UpdateOperation {
+			return warnings, fmt.Errorf("role type cannot be changed for an existing role")
```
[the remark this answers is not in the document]
pcman312: Done.

plugin/path_role.go
```diff
@@ -576,6 +617,19 @@ func (b *GcpAuthBackend) storeRole(ctx context.Context, s logical.Storage, roleN
 		return logical.ErrorResponse(err.Error()), nil
 	}
 
+	// Set default alias names
+	if role.IAMAliasType == "" {
+		role.IAMAliasType = defaultIAMAlias
+	}
+	if role.GCEAliasType == "" {
```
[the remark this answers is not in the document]
pcman312: This wouldn't be a backwards compatible change as described. An empty string needs to be interpreted as `unique_id` or `instance_id` otherwise the behavior will change for existing users. That being said, I'm okay with using a different value to not create an alias. Something like `noalias` or `no-alias`. I don't know the backstory as to why this is always creating an alias with no ability to not create an alias.

plugin/path_role_test.go
catsby: 👍

plugin/path_role.go
catsby: I don't necessarily disagree, however this breaks a common naming convention in most of our secret backends. I think `Role` is pretty Go idiomatic as well: 

- https://golang.org/doc/effective_go.html#Getters

pcman312: I agree with the naming convention for getters, however I don't agree that this is a getter as described. This is retrieving the role from the storage backend, not returning an existing, unexported field on the `GcpAuthBackend`.

plugin/aliasing.go
```diff
@@ -0,0 +1,97 @@
...
+	"fmt"
+	"sort"
+	"strconv"
+	"strings"
+
+	"google.golang.org/api/compute/v1"
+	"google.golang.org/api/iam/v1"
+)
+
+type aliasData struct {
```
broamski, line 13: This doesn't appear to be used anywhere?

plugin/aliasing.go
```diff
@@ -0,0 +1,97 @@
...
+		"":              getGCEInstanceID, // For backwards compatibility
+
+		"role_id": getGCERoleID,
+	}
+
+	allowedIAMAliasesSlice = iamMapKeyToSlice(allowedIAMAliases)
+	allowedGCEAliasesSlice = gceMapKeyToSlice(allowedGCEAliases)
+)
+
+func iamMapKeyToSlice(m map[string]iamAliaser) (s []string) {
```
broamski, line 45: These could probably be reduced into a single function that accepts an interface and is type switched. Same with many of the `get..` functions below.

plugin/path_role.go
catsby: That's fair, it's not a true getter as described in that link. I do feel that `role` is consistent with most of our other secret backends as well as some other methods/functions in this plugin (`Factory()`, `Backend()`, `IAMClient()`,`ComputeClient()`)

plugin/aliasing.go
pcman312: Removed. Leftover from a previous design that I scrapped.

plugin/aliasing.go
pcman312: Unfortunately it wouldn't really help here. The function would have the same code in it (including both versions of the for loop), but wrapped up into a single function rather than two. There isn't a generic `map` type that can be used as a variable type that would allow the type coercion needed here. Additionally it's generally frowned upon to pass the empty interface around unless necessary.

plugin/path_role.go
pcman312: Changed back to `role` for consistency across plugins that I wasn't aware of previously.

plugin/path_login.go
tyrannosaurus-becks: Marking this as resolved because we chatted outside this PR, and decided that in a follow-on PR well also let them configure the metadata.

plugin/path_role.go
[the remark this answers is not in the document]
kalafut: As far as I know, only the token auth method doesn't result in any alias. I'm not sure what the presumptions are of an alias always being set up either, though it would be worth a discussion about the use cases for no alias. It would be a significant change that we'd want to review, and would be larger than a minor update.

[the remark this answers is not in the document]
tyrannosaurus-becks: That's cool! Thanks for weighing in!

plugin/aliasing.go
tyrannosaurus-becks: I have generally found the same thing as well - when I have tried that approach it read awkwardly, but just in a different way. I don't think there's any "perfect" answer here, but I do like the more direct approach taken here because type conversions can be expensive.

kalafut:
@dustin-decker This change and another one to provide some customization of metadata storage (PR very soon) are targeted for 1.4.1.

PR #90, opened 2020-04-24
Allow auth metadata to be configured

tyrannosaurus-becks:
This draft PR shows the approach being currently pursued to add the ability to configure auth metadata at the role level.

Although it was added at the `identity/config` endpoint in a different auth engine, it was added at the role level in this one because that's where the aliases are also currently configured, so I thought it made sense to keep them together.

TODO
- [ ] Docs
- [ ] Wait for https://github.com/hashicorp/vault/pull/8783 to be merged and update to using Vault master

### DIFF
--- a/plugin/path_role.go
+++ b/plugin/path_role.go
@@ -9,6 +9,7 @@ import (
 	"github.com/hashicorp/go-gcp-common/gcputil"
 	"github.com/hashicorp/go-uuid"
 	"github.com/hashicorp/vault/sdk/framework"
+	"github.com/hashicorp/vault/sdk/helper/authmetadata"
 	vaultconsts "github.com/hashicorp/vault/sdk/helper/consts"
 	"github.com/hashicorp/vault/sdk/helper/tokenutil"
 	"github.com/hashicorp/vault/sdk/logical"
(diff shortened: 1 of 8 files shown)


plugin/path_role.go
```diff
@@ -37,6 +38,54 @@ const (
 	maxJwtExpMaxMinutes int = 60
 )
 
+var (
+	// The default gce_alias is "instance_id". The default fields
```
kalafut, line 42: For 1.4.1 I think we should choose defaults that match the 1.4 set since people will not be expecting that to change. AWS is somewhat different since that metadata was introduced quite recently and has known issues in some environments.

tyrannosaurus-becks: Sure, I can do that! Incorporating....

kalafut: I'll add a task for 1.5 to update these.

tyrannosaurus-becks: https://github.com/hashicorp/vault-plugin-auth-gcp/pull/90/commits/8d153f0a6e713a24d6abf55ac58777a53bcad734

kalafut:
@tyrannosaurus-becks @pcman312  I've been playing with this for both IAM and GCE and it seems to work fine. My main feedback is that we probably want this new configuration (both alias name and metadata) at the top level `/config` object, similar to AWS. (AWS has an additional "identity" level which doesn't apply here, but the configuration is still for the whole backend, not by role.)  Per-role configuration is a potential future enhancement.

tyrannosaurus-becks:
Deleting this branch because there are upstream conflicting changes, and the fields also need to be on the config, so it'll be faster to take off a fresh branch than continue with this one.

PR #92, opened 2020-04-27
GCP auth metadata config

tyrannosaurus-becks:
This PR adds the ability to configure the metadata that will be included on during login. The auth `Metadata` field drives what's written to audit logs. The auth _alias_ `Metadata` field drives what's included in Vault's identity system. 

A high rate of change to fields on alias metadata can cause performance issues at scale because each change triggers a storage write. The fields included in alias metadata here are identical to what was already previously included, and we simply add the ability to change them.
 
<details>
<summary>Command-line walk-through</summary>

```
$ vault auth enable -path=gcp vault-plugin-auth-gcp
Success! Enabled vault-plugin-auth-gcp auth method at: gcp/

$ vault read auth/gcp/config
Key             Value
---             -----
gce_metadata    [instance_creation_timestamp instance_id instance_name project_id project_number role service_account_id service_account_email zone]
iam_metadata    [project_id role service_account_id service_account_email]

$ vault write auth/gcp/config gce_metadata=""
Success! Data written to: auth/gcp/config

$ vault read auth/gcp/config
Key             Value
---             -----
gce_metadata    []
iam_metadata    [project_id role service_account_id service_account_email]

$ vault write auth/gcp/config gce_metadata="default"
Success! Data written to: auth/gcp/config

$ vault read auth/gcp/config
Key             Value
---             -----
gce_metadata    [instance_creation_timestamp instance_id instance_name project_id project_number role service_account_id service_account_email zone]
iam_metadata    [project_id role service_account_id service_account_email]

$ vault write auth/gcp/config gce_metadata="role"
Success! Data written to: auth/gcp/config

$ vault read auth/gcp/config
Key             Value
---             -----
gce_metadata    [role]
iam_metadata    [project_id role service_account_id service_account_email]

$ vault write auth/gcp/config gce_metadata="default,role"
Error writing data to auth/gcp/config: Error making API request.

URL: PUT http://localhost:8200/v1/auth/gcp/config
Code: 400. Errors:

* 1 error occurred:
	* failed to parse gce metadata: ["default" "role"] contains default - default can't be used in combination with other fields


$ vault write auth/gcp/config gce_metadata="badfield"
Error writing data to auth/gcp/config: Error making API request.

URL: PUT http://localhost:8200/v1/auth/gcp/config
Code: 400. Errors:

* 1 error occurred:
	* failed to parse gce metadata: "badfield" contains an unavailable field, please select from "instance_creation_timestamp, instance_id, instance_name, project_id, project_number, role, service_account_id, service_account_email, zone"
```

</details>

<details>
<summary>Test output</summary>

```
$ make test
ok  	github.com/hashicorp/vault-plugin-auth-gcp/plugin	(cached)
```

</details>

TODO
- [ ] Write docs (separate PR)


plugin/gcp_config.go
```diff
@@ -4,6 +4,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"github.com/hashicorp/vault/sdk/helper/authmetadata"
```
pcman312, line 7: nit: import ordering? I think `goimports` fixes this

plugin/path_config.go
```diff
@@ -133,18 +165,18 @@ func (b *GcpAuthBackend) pathConfigRead(ctx context.Context, req *logical.Reques
...
 		return nil, err
 	}
-	if entry == nil {
-		return nil, nil
-	}
-
-	if err := entry.DecodeJSON(config); err != nil {
-		return nil, err
+	if entry != nil {
+		if err := entry.DecodeJSON(config); err != nil {
```
pcman312, line 177: Why move this into a nested structure?

plugin/path_config_test.go
```diff
@@ -31,9 +33,17 @@ func TestBackend_PathConfigRead(t *testing.T) {
 		if err != nil {
 			t.Fatal(err)
 		}
-
-		if resp != nil {
-			t.Errorf("expected response data to be empty, got %v", resp.Data)
+		if resp == nil {
```
pcman312, line 36: Can an expected response object be created and then use `reflect.DeepEqual` instead of these checks?

plugin/path_login.go
```diff
@@ -499,11 +500,12 @@ func (b *GcpAuthBackend) pathGceLogin(ctx context.Context, req *logical.Request,
 		Alias: &logical.Alias{
 			Name: alias,
 		},
-		Metadata:    authMetadata(loginInfo, serviceAccount),
 		DisplayName: instance.Name,
 	}
-
 	role.PopulateTokenAuth(auth)
+	if err := conf.GCEAuthMetadata.PopulateDesiredMetadata(auth, authMetadata(loginInfo, serviceAccount)); err != nil {
+		b.Logger().Warn(fmt.Sprintf("unable to populate gce metadata due to %s", err))
```
pcman312, line 507: `b.Logger().Warn("unable to populate GCE metadata", "error", err.Error())`?

plugin/path_login.go
```diff
@@ -346,11 +346,12 @@ func (b *GcpAuthBackend) pathIamLogin(ctx context.Context, req *logical.Request,
 		Alias: &logical.Alias{
 			Name: alias,
 		},
-		Metadata:    authMetadata(loginInfo, serviceAccount),
 		DisplayName: serviceAccount.Email,
 	}
-
 	role.PopulateTokenAuth(auth)
+	if err := conf.IAMAuthMetadata.PopulateDesiredMetadata(auth, authMetadata(loginInfo, serviceAccount)); err != nil {
+		b.Logger().Warn(fmt.Sprintf("unable to populate iam metadata due to %s", err))
```
pcman312, line 353: `b.Logger().Warn("unable to populate IAM metadata", "error", err.Error())`?

plugin/path_config_test.go
```diff
@@ -318,15 +350,10 @@ func TestConfig_Update(t *testing.T) {
 				tc.d.Schema = pathConfig(&b).Fields
 			}
 
-			changed, err := tc.new.Update(tc.d)
-			if (err != nil) != tc.err {
+			if err := tc.new.Update(tc.d); (err != nil) != tc.err {
```
pcman312, line 353: This is really difficult to read. Can it be simplified?
```go
err := tc.new.Update(tc.d)
if tc.err && err == nil {
    t.Fatalf("err expected, got nil")
}
if !tc.err && err != nil {
    t.Fatalf("no error expected, got: %s", err)
}
```

plugin/path_config_test.go
```diff
@@ -2,6 +2,8 @@ package gcpauth
 
 import (
 	"context"
+	"fmt"
+	"github.com/hashicorp/vault/sdk/helper/authmetadata"
```
pcman312, line 6: nit: import ordering

plugin/gcp_config.go
tyrannosaurus-becks: Fixed the imports but I won't link the commit since I'm going to squash them all before I mark this ready as a PR.

plugin/path_config_test.go
tyrannosaurus-becks: Fixed the imports but I won't link the commit since I'm going to squash them all before I mark this ready as a PR.

plugin/path_config.go
```diff
@@ -64,36 +102,28 @@ func (b *GcpAuthBackend) pathConfigWrite(ctx context.Context, req *logical.Reque
...
 	if err != nil {
 		return nil, err
 	}
-	if c == nil {
-		c = &gcpConfig{}
-	}
 
-	changed, err := c.Update(d)
-	if err != nil {
+	if err := c.Update(d); err != nil {
```
tyrannosaurus-becks, line 109: I stripped the `changed` field being returned because when auth metadata fields from the user are parsed, the SDK doesn't return whether it's changed. So, there was a choice of either adding that to the SDK, or stripping the field and always updating the config. I went with stripping the field because 1) no other plugin I've seen only stores things if a field changed, and 2) if people are creating or updating a config, it's probably safe to assume it's intended to change.

plugin/path_config.go
```diff
@@ -64,36 +102,28 @@ func (b *GcpAuthBackend) pathConfigWrite(ctx context.Context, req *logical.Reque
 	}
 
 	c, err := b.config(ctx, req.Storage)
-
 	if err != nil {
 		return nil, err
 	}
-	if c == nil {
```
tyrannosaurus-becks, line 71: I stripped all the checks for whether the config is nil because I made it so the config is _always_ returned by the `b.config(ctx, req.Storage)` method. I did that because even if they haven't configured anything, they effectively _still_ have a config because the metadata fields are defaulting. This is consistent with the approach taken with AWS.

plugin/path_config.go
tyrannosaurus-becks: I did it because previously, if the entry was nil, it returned so we never needed to worry about a panic on this line. However, on this branch I edited a config to always be returned so that the config would always have the new handlers on it, avoiding any risk of panics.

pcman312: I recommend inverting the conditional it as it makes it easier to follow and keeps the happy path towards the left:
```go
if entry == nil {
  return config, nil
}
err = entry.DecodeJSON(config)
if err != nil {
  return nil, err
}
return config, nil
```

plugin/path_config.go
pcman312: Is this a safe assumption to make? I'm looking through to double check but I just want to be paranoid about this kind of change.

plugin/path_config.go
tyrannosaurus-becks: Can do! https://github.com/hashicorp/vault-plugin-auth-gcp/pull/92/commits/d96be7888a88a611ffb68f1bdcd20f09386c9837

plugin/path_login.go
```diff
@@ -505,11 +500,12 @@ func (b *GcpAuthBackend) pathGceLogin(ctx context.Context, req *logical.Request,
 		Alias: &logical.Alias{
 			Name: alias,
 		},
-		Metadata:    authMetadata(loginInfo, serviceAccount),
 		DisplayName: instance.Name,
 	}
-
 	role.PopulateTokenAuth(auth)
+	if err := conf.GCEAuthMetadata.PopulateDesiredMetadata(auth, authMetadata(loginInfo, serviceAccount)); err != nil {
+		b.Logger().Warn("unable to populate gce metadata due", "err", err.Error())
```
austingebauer, line 507: Nit: suggest to remove `due` at the end of this error message so that the message is similar to the IAM auth error message above.

tyrannosaurus-becks: Good eye! Will fix!

tyrannosaurus-becks: https://github.com/hashicorp/vault-plugin-auth-gcp/pull/92/commits/0dcbae999030110ed011ac6f89ba6c2992def536

plugin/path_config.go
tyrannosaurus-becks: Yes, I think it's good to be extra safe.

Just to be sure, I waited until I was fresh this morning and went through everywhere it's used again. I'm going to go through this PR and comment upon each location. 

For this location, it actually looks like [this would have panicked previously](https://github.com/hashicorp/vault-plugin-auth-gcp/blob/d0062d35cf3848e2e912a0d6f4afa214390b8bf4/plugin/backend.go#L174) if the config was nil because it's immediately used below. So, this code represents a behavior change. In looking at the code now, I think it actually fixes that bug and now does what's intended, which is to continue on to the default credentials below. So, this should be a safe change and actually have a positive effect.

plugin/path_config.go
```diff
@@ -106,23 +136,25 @@ func (b *GcpAuthBackend) pathConfigRead(ctx context.Context, req *logical.Reques
 	if err != nil {
 		return nil, err
 	}
-	if config == nil {
```
tyrannosaurus-becks, line 109: For this location, the effect of the config never being nil is desirable because now reads when the config is unset will result in the user learning their default alias, and their default auth metadata.

plugin/path_config.go
tyrannosaurus-becks: For this location, the effect of the config never being nil is basically no change since it was created below if it was nil.

pcman312: `formatAndMarshalCredentials` is nil safe: https://github.com/hashicorp/vault-plugin-auth-gcp/blob/master/plugin/path_config.go#L145

plugin/path_login.go
```diff
@@ -320,9 +320,6 @@ func (b *GcpAuthBackend) pathIamLogin(ctx context.Context, req *logical.Request,
 	if err != nil {
 		return logical.ErrorResponse("unable to retrieve GCP configuration"), nil
 	}
-	if conf == nil {
```
tyrannosaurus-becks, line 323: For this location, the effect of the config being unset but populated is basically no change from before we did any work on this plugin. Before, the config was mainly used in building the client, and if the client had no available credentials, it would fail there. It still will with this approach.

plugin/path_login.go
```diff
@@ -468,9 +466,6 @@ func (b *GcpAuthBackend) pathGceLogin(ctx context.Context, req *logical.Request,
...
```
tyrannosaurus-becks, line 471: Same explanation as path_login.go comment above.

plugin/path_config.go
tyrannosaurus-becks: Also, having the returned config be non-nil solves a bug [here](https://github.com/hashicorp/vault-plugin-auth-gcp/blob/d0062d35cf3848e2e912a0d6f4afa214390b8bf4/plugin/backend.go#L174) where it would have panicked if the config were nil. Now it will continue on to using the default credentials as the code intends.

plugin/path_config_test.go
pcman312: Will this work instead of multiple separate assertions?
```go
expectedResp := &logical.Response{
  Data: map[string]interface{}{
    "iam_metadata": []string{
      ...
    },
    "gce_metadata": []string{
      ...
    },
  },
}

if !reflect.DeepEqual(resp, expectedResp) {
  t.Fatalf("Actual: %#v\nExpected: %#v", resp, expectedResp)
}
```

tyrannosaurus-becks: Sure, pushed in https://github.com/hashicorp/vault-plugin-auth-gcp/pull/92/commits/2495f0ee52f7997018c9e11e1cf53bacd2a328aa

PR #98, opened 2020-06-23
Adding label groups and blocklist label groups

soprea:
# Overview

This change will allow entities to join vault groups based on labels.
With this you can use policy templates for easier access management. 
If not in use, no change on existing configuration is required.
The config fields introduced are not mandatory.
They don't break existing configs or workflow - completely backwards compatible.

# Design of Change
If you set `add_group_label_aliases` to `true`, the response will include the values of the key/value pair from instance labels to the Authentication Group Aliases response and Vault will add the entity to the external vault group that was created for the value of the label.
The Group Alias will need to be setup as an external group alias with the gcp auth method as mount.

[Example]
GCP Labels: `environment:dev, app_name:vault`.
The instances with this labels will have their vault entities added to the `dev` and `vault` external groups aliases.

If you set `blocklist_group_label_aliases` to a list of strings, the auth method will not add the value of the key from instances labels with given name to the response. - This will result in a remove from group if the entity was already added (from a previous auth/config).

`blocklist_group_label_aliases` works only with `add_group_label_aliases`. If `blocklist_group_label_aliases` is set but `add_group_label_aliases` is set to false or not set at all - nothing happens

[Example]
GCP Labels: `environment:dev, app_name:vault` and `blocklist_group_label_aliases = "environment"`, the instances with this labels will have their vault entities added to the  `vault` external group only.

 # Tests
Acceptance tests passed.

Tested on our lab as well on different scenarios: with a blocklist, without blocklist (with empty blocklist), with add_group label aliases and without.

# Notes

No PR against main vault website. I want to have the thumbs up from you guys first.
For anything else, please don't hesitate - I'm fairly new to this things :)


plugin/path_login.go
```diff
@@ -511,15 +511,42 @@ func (b *GcpAuthBackend) pathGceLogin(ctx context.Context, req *logical.Request,
...
-			return nil, err
+		if role.AddGroupLabelAliases {
+			labels, err := b.groupLabelAliases(computeClient, ctx, metadata.ProjectId, metadata.Zone, metadata.InstanceName)
+			if err != nil {
+				return nil, err
+			}
+			for k, v := range labels {
+				if len(role.BlocklistGroupLabelAliases) > 0 {
+					for _, blocklistLabel := range role.BlocklistGroupLabelAliases {
+						if k != blocklistLabel {
```
Valarissa, line 536: I believe that this behavior is incorrect. I think you want to create a flag that is initialized as false and iterates over the `BlocklistGroupLabelAliases` and sets it to true if the label is found in the blocklist. If at the end of that iteration, the value is false, then you add the alias.

As it stands, I believe this will add the alias multiple times for every case that the label is not a particular blocklist entry.

Valarissa (changes requested):
I found one substantial bug in this PR.

In addition, it would also be good to ensure that we have a test that actually exercises this code more than just the option parsing. You might be able to mock the call to `groupLabelAliases` to return something without actually interacting with GCP.

It would also be good to ensure that your workflows and the automated tests test this when there's more than one label in the blocklist as well.

plugin/path_login.go
calvn: @Valarissa is correct. We should only be appending once we've checked the entire slice.

PR #99, opened 2020-07-16
Changes the default alias names to role_id for both IAM and GCE authn

austingebauer:
# Overview

This PR changes the default name of the entity alias that gets created for both IAM and GCE authentication. Test fixes are also included in this PR. See test output below.

For IAM, changes it from `unique_id ` to `role_id`.
For GCE, change it from `instance_id ` to `role_id`.

Docs PR: https://github.com/hashicorp/vault/pull/9494

# Design of Change

Change was implemented using alias configuration made available in https://github.com/hashicorp/vault-plugin-auth-gcp/pull/89.

- https://github.com/hashicorp/vault-plugin-auth-gcp/pull/89

# Contributor Checklist
- [x] Add relevant docs to upstream Vault repository, or sufficient reasoning why docs won’t be added yet
- [ ] Backwards compatible
    - This is an intentional change that will be included in a CHANGES entry in the Vault CHANGELOG.
- [x] Add output for any tests not ran in CI to the PR description (eg, acceptance tests)

```
$ export GOOGLE_CREDENTIALS="$(cat /Users/austingebauer/.gcp/austin-gebauer-db62ff1c96b4.json)"
$ make test                                                                                    
?       github.com/hashicorp/vault-plugin-auth-gcp      [no test files]
ok      github.com/hashicorp/vault-plugin-auth-gcp/plugin       (cached)
?       github.com/hashicorp/vault-plugin-auth-gcp/plugin/cache [no test files]
$ echo $?
0
```

### DIFF
--- a/plugin/aliasing.go
+++ b/plugin/aliasing.go
@@ -13,22 +13,22 @@ type iamAliaser func(role *gcpRole, svcAccount *iam.ServiceAccount) (alias strin
 type gceAliaser func(role *gcpRole, instance *compute.Instance) (alias string)
 
 const (
-	defaultIAMAlias = "unique_id"
-	defaultGCEAlias = "instance_id"
+	defaultIAMAlias = "role_id"
+	defaultGCEAlias = "role_id"
 )
 
 var (
 	allowedIAMAliases = map[string]iamAliaser{
-		defaultIAMAlias: getIAMSvcAccountUniqueID,
-		"":              getIAMSvcAccountUniqueID, // For backwards compatibility
+		defaultIAMAlias: getIAMRoleID,
+		"":              getIAMRoleID, // For backwards compatibility
 
-		"role_id": getIAMRoleID,
+		"unique_id": getIAMSvcAccountUniqueID,
 	}
 	allowedGCEAliases = map[string]gceAliaser{
-		defaultGCEAlias: getGCEInstanceID,
-		"":              getGCEInstanceID, // For backwards compatibility
+		defaultGCEAlias: getGCERoleID,
+		"":              getGCERoleID, // For backwards compatibility
 
-		"role_id": getGCERoleID,
+		"instance_id": getGCEInstanceID,
 	}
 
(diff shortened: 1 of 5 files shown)


plugin/aliasing.go
```diff
@@ -13,22 +13,22 @@ type iamAliaser func(role *gcpRole, svcAccount *iam.ServiceAccount) (alias strin
...
+	defaultIAMAlias = "role_id"
+	defaultGCEAlias = "role_id"
 )
 
 var (
 	allowedIAMAliases = map[string]iamAliaser{
-		defaultIAMAlias: getIAMSvcAccountUniqueID,
-		"":              getIAMSvcAccountUniqueID, // For backwards compatibility
+		defaultIAMAlias: getIAMRoleID,
+		"":              getIAMRoleID, // For backwards compatibility
```
austingebauer, line 23: I'm not completely sure that this `""` entry is needed for backwards compatibility anymore now that the default is changing. @pcman312, do you have any context on this?

pcman312: The empty string is there to allow for a value to not be set -> default value. I think that this should remain here and default to the role_id as you have it.

austingebauer: That makes sense. I'm just wondering if it's ever used given the logic at https://github.com/hashicorp/vault-plugin-auth-gcp/blob/master/plugin/gcp_config.go#L94. `""` gets replaced with the default key.

pcman312: Probably not - If I recall correctly I put the empty string check in as an over-abundance of caution. If we want to remove one, I'd prefer to remove the empty string check so the map is a single place where we see all of the values + default together.

austingebauer: Got it 👍 . Done in 74e03c93fa7577ea9caa1d6fb9127ed7ae83a811.

austingebauer: Reran tests successfully as well.

PR #103, opened 2020-10-05
String was being passed in sprintf format instead of as Wrapf param

MitchellBot:
# Overview
Super minor change. We're having an issue with auth that is likely environmental but the error we're getting is `* unable to get public key for signed JWT: %v`. It would be much more helpful if we received the actual error instead of %v. 

# Design of Change
Verified that errwrap.Wrapf replaces {{err}}, not %v and updated call to match.

No related issue, simply something I noticed while troubleshooting an environmental issue.

# Contributor Checklist
No tests were run, frankly, I didn't do any verification except to check that Wrapf is expecting {{err}} which you can see in the function itself and just a few lines before the changed line in another call to Wrapf

### DIFF
--- a/plugin/path_login.go
+++ b/plugin/path_login.go
@@ -173,7 +173,7 @@ func (b *GcpAuthBackend) parseAndValidateJwt(ctx context.Context, req *logical.R
 
 	key, err := b.getSigningKey(ctx, jwtVal, signedJwt.(string), loginInfo.Role, req.Storage)
 	if err != nil {
-		return nil, errwrap.Wrapf("unable to get public key for signed JWT: %v", err)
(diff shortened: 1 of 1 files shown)


kalafut:
This seems fine pending the CLA signing.

kalafut:
@MitchellBot The check is pretty quick (< 1m). If the email(s) in any of the commits are not in your Github profile, it won't pass.

James-Mitchell-SYMC:
@kalafut Figured it out. Somehow the commit was under an old, defunct account.

kalafut (approved):

kalafut:
Thanks!

PR #110, opened 2021-04-29
Update plugin-backport workflow to work with label events

catsby:
This PR changes our installed backporter workflow to also run when merged PRs are labeled.


tomhjp (approved):
LGTM. For any PRs that are already labelled with backport labels, I think you will need to remove and re-add the label.

Issue #85, opened 2020-03-26
Add support for GKE Metadata server


schuylr:
Google Cloud has recently started including Workload Identity into GKE. As a result, node pools with GKE Metadata server enabled no longer have the GCE metadata associated with it.

See [here](https://cloud.google.com/kubernetes-engine/docs/how-to/workload-identity#using_workload_identity_from_your_code) for more information and this screenshot regarding the setting: 

![image](https://user-images.githubusercontent.com/12628715/77703038-3138c800-6f90-11ea-8987-8f3f76e138ad.png)

Is there any way that the GCE implementation can work with the new GKE metadata servers?

pperez:
Would love to get this working, we love workload identity!

gysel:
This this related?

https://discuss.hashicorp.com/t/vault-agent-gcp-auto-auth-gke-workload-identity/10518

PR #115, opened 2021-08-02
Add support for autogenerating GCE/GKE id_token

upodroid:
Right now, gce login method requires the enduser to mint a jwt manually by running the following from https://www.vaultproject.io/docs/auth/gcp#gce

```
ROLE="my-gce-role"

curl \
  --header "Metadata-Flavor: Google" \
  --get \
  --data-urlencode "audience=http://vault/${ROLE}" \
  --data-urlencode "format=full" \
  "http://metadata/computeMetadata/v1/instance/service-accounts/default/identity"

```

My change will allow the following to work correctly.

```
vault login -method=gcp \
    role=$VAULT_ROLE
```

This allows the following behavior:
- Autodiscovery of service account id_token on Google Cloud
- Support for GKE Workload Identity.
- Support for running vault auth from Cloud Run

Old behavior is mantained by specifying service_account parameter.

You can verify the change with the following code snippet:
```
// main.go
package main
import (
        "cloud.google.com/go/compute/metadata"
        "net/http"
        "net/url"
        "fmt"
)
func main() {
        if metadata.OnGCE() {
                fmt.Println("We are running on GCP")
                metadataClient := metadata.NewClient(&http.Client{})
                v := url.Values{}
                v.Set("audience", fmt.Sprintf("http://vault/%s", "fos"))
                v.Set("format", "full")
                path := "instance/service-accounts/default/identity?" + v.Encode()
                instanceJwt, _ := metadataClient.Get(path)
                fmt.Println(instanceJwt)
        }
}
```
Fixes: https://github.com/hashicorp/vault-plugin-auth-gcp/issues/85

# Contributor Checklist
[X] Add relevant docs to upstream Vault repository, or sufficient reasoning why docs won’t be added yet
[My Docs PR Link](https://github.com/hashicorp/vault/pull/12231)
[X] Add output for any tests not ran in CI to the PR description (eg, acceptance tests)
[X] Backwards compatible


upodroid:
@kalafut @sethvargo @salrashid123 Can you review this?

Is it me or is this plugin a lot more convulted than it needs to be?

https://www.vaultproject.io/docs/auth/gcp

If you are on Google Cloud, you should just read the identity token of the GCE Instance/K8s Pod/GCF/Cloudrun and present it to vault to validate.
If you are outside of Google Cloud, you should present a JWT signed as described in the docs.

upodroid:
Also, the users with Google Cloud service account can simply run 
`gcloud auth print-identity-token --account vault-dev@REDACTED.gserviceaccount.com --audiences http://vault/role` to generate an id_token instead of having to deal with IAMServiceAccounts API. 

https://pkg.go.dev/google.golang.org/api/idtoken#NewTokenSource can do the heavy lifting instead of that API.

salrashid123:
@upodroid   i'm not a maintainer on this repo but can state that while on compute, getting an id_token via impersonation/iam api should not be necessary since you can "just get it" via metadata:  
  https://github.com/salrashid123/vault_gcp#vault-auth-for-gce-instances

---

just fyi on the topic of oidc and impersonation with _secrets_ (not vault auth), see link [here](https://github.com/salrashid123/vault-plugin-secrets-gcp) << not that is an unsupported fork.  if interested in those flows, it'd need to be in a supported branch here (i have a cople of feature requests for those)

plugin/cli.go
```diff
@@ -33,44 +35,54 @@ func getSignedJwt(role string, m map[string]string) (string, error) {
 		serviceAccount = credentials.ClientEmail
 	}
 	if serviceAccount == "" {
+		// Check if the metadata server is available.
+		if metadata.OnGCE() {
+			metadataClient := metadata.NewClient(&http.Client{})
+			path := "instance/service-accounts/default/identity?audience=http%3A%2F%2Fvault%2F" + role + "&format=full"
```
sethvargo, line 41: I think we need a bit more rigor around building this path. In particular, the query escaping might get weird if the role name has special characters. It would probably be safer to build a [`url.Values`](https://pkg.go.dev/net/url#Values) instead.

plugin/cli.go
```diff
@@ -33,44 +35,54 @@ func getSignedJwt(role string, m map[string]string) (string, error) {
...
+			instanceJwt, err := metadataClient.Get(path)
+			if err != nil {
+				return "", err
```
sethvargo, line 44: Wrap this error

go.sum
```diff
@@ -1,36 +1,97 @@
 cloud.google.com/go v0.26.0/go.mod h1:aQUYkXzVsufM+DwF1aE+0xfcU+56JwCaLick0ClmMTw=
```
sethvargo, line 1: Can you run `go mod tidy` and `go mod verify` please?

sethvargo (commented):
@upodroid - I left a few comments on the PR. I'm a little worried about the way the metadata path is currently constructed, but otherwise this seems reasonable.

> Is it me or is this plugin a lot more convoluted than it needs to be?

You're both correct and incorrect 😄 . This plugin predates a lot of the functionality you've mentioned. To name a few: Cloud Run didn't event exist yet and Workload Identity Federation would come almost 2 years later. This plugin also predates the IAMCredentials API by about a year.

I wrote up a proposal for modernizing the plugin (there's also some Vault changes that dramatically reduce the complexity of the code here) which could warrant a v2 of the plugin the future.

go.sum
upodroid:
```
 REDACTED  MCW0CDP3YY  ~  go  …  github.com  hashicorp  vault-plugin-auth-gcp   instance-jwt-auth  $    go mod tidy
 REDACTED  MCW0CDP3YY  ~  go  …  github.com  hashicorp  vault-plugin-auth-gcp   instance-jwt-auth  $    go mod verify
all modules verified
 REDACTED  MCW0CDP3YY  ~  go  …  github.com  hashicorp  vault-plugin-auth-gcp   instance-jwt-auth  $    git diff
 REDACTED  MCW0CDP3YY  ~  go  …  github.com  hashicorp  vault-plugin-auth-gcp   instance-jwt-auth  $    
```

Didn't generate a diff.

plugin/cli.go
upodroid: Fixed

plugin/cli.go
upodroid: Done

plugin/cli.go
```diff
@@ -33,44 +36,57 @@ func getSignedJwt(role string, m map[string]string) (string, error) {
...
+		// Check if the metadata server is available.
+		if metadata.OnGCE() {
+			metadataClient := metadata.NewClient(&http.Client{})
+			v := url.Values{}
+			v.Set("audience", fmt.Sprintf("http://vault/%s", role))
+			v.Set("format", "full")
+			path := "instance/service-accounts/default/identity?" + v.Encode()
+			instanceJwt, err := metadataClient.Get(path)
+			if err != nil {
+				return "", fmt.Errorf("unable to read the identity token: %v", err)
```
sethvargo, line 48:
```suggestion
				return "", fmt.Errorf("unable to read the identity token: %w", err)
```

plugin/cli.go
```diff
@@ -33,44 +36,57 @@ func getSignedJwt(role string, m map[string]string) (string, error) {
 		serviceAccount = credentials.ClientEmail
 	}
 	if serviceAccount == "" {
+		// Check if the metadata server is available.
+		if metadata.OnGCE() {
```
sethvargo, line 40: This is _super_ nitty, but it could be a bit cleaner if you negated this (saves on some indentation). For example:

```go
if !metadata.OnGCE() {
  return "", errors.New("could not obtain...")
}

// Metadata server is available, pull creds from there
metadataClient := metadata.NewClient(&http.Client{})
...
```

plugin/cli.go
```diff
@@ -33,44 +36,57 @@ func getSignedJwt(role string, m map[string]string) (string, error) {
...
+			metadataClient := metadata.NewClient(&http.Client{})
```
sethvargo, line 41:
```suggestion
			metadataClient := metadata.NewClient(cleanhttp. DefaultClient())
```

plugin/cli.go
upodroid: Fixed

upodroid:
After this PR, i'm open to rewriting the plugin to support the following google authentication methods:
- idToken read from a GCE instance using the gce role and it's bound parameters
- idToken read from GKE/Cloudrun to work with iam method
- idToken generated from Service Accounts 
- idToken generated by the IAM Serviceaccounts API from an impersonated service account.
- Reject idTokens from anywhere else including user ADCs
- Rewrite the docs to support these 4 methods only.


Do you have a link to that proposal?

pratheekrebala:
FWIW, It seems like some of this logic is already included in the Vault core when using the Vault Agent's `auto_auth` method.

https://github.com/hashicorp/vault/blob/14c28ceefe364e7610b74f4f057f2bed35f4cd9e/command/agent/auth/gcp/gcp.go#L128

voxsim:
@upodroid @sethvargo do you have any news on this PR? Thank you 🙏

hsimon-hashicorp:
@upodroid Can you resolve the go.mod and go.sum conflicts? I'll have @imthaghost take a peek at it too. Thanks!

hsimon-hashicorp:
And one more task request - if you could please add a changelog entry, we'll need that too. :) Thanks!!

upodroid:
Workload Identity Federation should work as that feature is implemented in the golang Google Libraries. Try it and see how it goes.

Please authenticate using the Google ADC(Application Default Credentials) to ensure Workload Identity Fed works properly.

ekhaydarov:
@upodroid I have tried implementing it with the `iam` and `gcp` config type. I get the same error of credentials being of different type, map not string. I may be configuring it incorrectly so will take a look another time but it doesnt seem straight forward. 

I do not want to pollute this thread with a help request so apologies. I wonder if I can talk with someone in the google cloud community slack about this or should I simply raise a new issue?

robmonte:
Hi @upodroid Thanks for your contribution! We want to take a closer look at both the PR and feedback in this discussion, so I'm going to get it prioritized in our internal tracker.

robmonte:
Hello again @upodroid 
I only have 2 small comments after reviewing the PR, which are above. If you could check them out, we can get this merged in soon - thanks again!

upodroid:
Thanks for getting back to me. 

- The changelog has been added.
- Pushed a commit to fix go.mod conflicts

robmonte:
Sorry, forgot to hit submit on the review comments 😅

plugin/cli.go
```diff
@@ -33,44 +35,58 @@ func getSignedJwt(role string, m map[string]string) (string, error) {
...
-		return "", fmt.Errorf("could not convert JWT payload to JSON string: %v", err)
-	}
+		jwtReq := &iamcredentials.SignJwtRequest{
+			Payload: string(payloadBytes),
+		}
 
-	jwtReq := &iamcredentials.SignJwtRequest{
-		Payload: string(payloadBytes),
-	}
+		iamClient, err := iamcredentials.New(httpClient)
```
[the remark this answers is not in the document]
upodroid: Fixed.

robmonte (approved):

PR #129, opened 2022-04-05
Fix changelog order and downgrade google.golang.org/api

robmonte:
Move 0.11.3 bug fix below unreleased features in the changelog file


calvn (approved):
LGTM, and it seems that were back to the state that we were on prior to the upgrade: https://github.com/hashicorp/vault-plugin-auth-gcp/blob/a21b9f510ae24d100252073bf1991378d61d20e0/go.mod

austingebauer (approved):
👍

Issue #131, opened 2022-06-16
Reset vault gcp auth to Application Default Credentials


ryanhulet:
I cannot find a way to set the gcp/config to empty. 

If I do 
`vault write gcp/config credentials="{}"` 
I get the following error: 
`* could not create IAM Admin client: failed to create IAM HTTP client: failed to create oauth2 http client: failed to parse credentials: missing 'type' field in credentials`

and
`vault write gcp/config credentials=""`
errors to
`* invalid credentials JSON file: unexpected end of JSON input`

austingebauer:
Hi, @ryanhulet. Thanks for reporting this. I _think_ you're correct that there isn't currently a way to reset the credentials parameter on the config. I'm going to look closer at this to see if I can find a way.

ryanhulet:
Essentially I am trying to move from using defined credentials to the env variable GOOGLE_APPLICATION_CREDENTIALS without destroying and rebuilding our vault cluster

austingebauer:
@ryanhulet - That makes sense! Currently, I don't see a way to do that. I'm going to experiment on this to find a potential workaround. In any case, I think this is worth fixing up to make it easier. Thanks for reporting it.

PR #132, opened 2022-06-29
Fix config reset to application default credentials

austingebauer:
## Overview

This PR fixes the ability to reset the configuration's credentials to use application default credentials. Prior to this change, it wasn't possible to update the config to begin using application default credentials again.

Fixes https://github.com/hashicorp/vault-plugin-auth-gcp/issues/131.

## Testing

I've added a couple of test cases to this PR. I also successfully ran all acceptance tests.

### DIFF
--- a/plugin/gcp_config.go
+++ b/plugin/gcp_config.go
@@ -61,16 +61,24 @@ func (c *gcpConfig) Update(d *framework.FieldData) error {
 	}
 
 	if v, ok := d.GetOk("credentials"); ok {
-		creds, err := gcputil.Credentials(v.(string))
-		if err != nil {
-			return fmt.Errorf("failed to read credentials: %w", err)
-		}
+		credentials := v.(string)
 
-		if len(creds.PrivateKeyId) == 0 {
-			return errors.New("missing private key in credentials")
-		}
+		// If the given credentials are empty, reset them so that application default
+		// credentials are used. Otherwise, parse and validate the given credentials.
+		if credentials == "" {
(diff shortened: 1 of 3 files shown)


plugin/gcp_config.go
```diff
@@ -61,16 +61,24 @@ func (c *gcpConfig) Update(d *framework.FieldData) error {
...
-		if err != nil {
-			return fmt.Errorf("failed to read credentials: %w", err)
-		}
+		credentials := v.(string)
+
+		// If the given credentials are empty, reset them so that application default
+		// credentials are used. Otherwise, parse and validate the given credentials.
+		if credentials == "" {
+			c.Credentials = nil
+		} else {
```
robmonte, line 70: Overall looks good to me.

I have a question about the logic this is based on, if you know. It looks to me like [this line](https://github.com/hashicorp/vault-plugin-auth-gcp/blob/a6717f99edfbe8c64517529a7a66f1f68962ce4a/plugin/backend.go#L213=) is where it's deciding whether to use Application Default Credentials or the credentials provided inside of the config. How come overriding credentials to `""`, before your PR, does not satisfy `len(credBytes)` to be 0 and cause it to enter the else block?

robmonte: Thanks for the explanation, I see it now.

robmonte (approved):
Looks good! Makes me wonder if we have any other plugins with fields that act in a similar manner and cannot be reset.

calvn (approved):

PR #135, opened 2022-07-15
VAULT-6727 role resolution for GCP

VioletHynes:
Support role resolution functionality (to work with role based quotas). The intent of this operation is to take the same data as a login request and return the role that would be used in the request, so that we can quota upon login requests that have a certain role.

Note that the mocks_test change was required as part of the SDK update, but has no impact.

Afterwards, I will push a new tag and pull this new version into Vault.


hghaf099 (approved):
lgtm

plugin/path_login.go
```diff
@@ -141,6 +156,22 @@ type gcpLoginInfo struct {
...
+	if roleName == "" {
+		return "", nil, errors.New("role is required")
+	}
+
+	role, err := b.role(ctx, s, roleName)
+	if err != nil {
+		return "", nil, err
+	}
+	if role == nil {
+		return "", nil, fmt.Errorf("role '%s' not found", roleName)
```
ccapurso, line 170: nit: You can make use of `%q` instead of `'%s'` here.

ccapurso (approved):
One tiny nit but otherwise looks good!

plugin/path_login.go
tomhjp: I think it's better to update and improve instead of keeping bad patterns, please do feel free to change this to `%q`

tomhjp (approved):
LGTM

CHANGELOG.md
```diff
@@ -7,6 +7,7 @@ BUG FIXES:
 IMPROVEMENTS:
 
 * Updates dependencies: `google.golang.org/api@v0.83.0`, `github.com/hashicorp/go-gcp-common@v0.8.0` [[GH-130](https://github.com/hashicorp/vault-plugin-auth-gcp/pull/130)]
+* Enables GCP roles to be compatible with Vault's role based quotas [[GH-135](https://github.com/hashicorp/vault-plugin-auth-gcp/pull/130)].
```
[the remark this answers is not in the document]
VioletHynes: Good spot - thanks :)

PR #194, opened 2024-02-05
refactor plugin to use release tooling

kschoche:
moves `main.go` to `cmd/` so the release tooling can build, tag, and release automatically.


swenson (approved):
LGTM so far

Makefile
```diff
@@ -22,7 +22,7 @@ default: dev
 # is only put into /bin/
 .PHONY: quickdev
 quickdev: generate
-	@CGO_ENABLED=0 go build -i -tags='$(BUILD_TAGS)' -o bin/vault-gcp-auth-plugin
+	@CGO_ENABLED=0 go build -i -tags='$(BUILD_TAGS)' -o bin/vault-plugin-auth-gcp cmd/vault-plugin-auth-gcp/main.go
```
benashz, line 25: Could use `$(TOOL)` here, instead of repeating it?

benashz (approved):
One minor nit to take a look at. Otherwise 👍

Makefile
kschoche: Good call, also apparently `-i` flag is invalid now:
```
demo $ make quickdev
flag provided but not defined: -i
usage: go build [-o output] [build flags] [packages]
```
So I removed that as well.
Fixed in a66468f

PR #199, opened 2024-04-01
Automated dependency upgrades

hc-github-team-secure-vault-ecosystem:
Full log: https://github.com/hashicorp/vault-plugin-auth-gcp/actions/runs/8707230961

  IMPROVEMENTS:
* Updated dependencies:
   * `cloud.google.com/go/compute/metadata` v0.2.3 -> v0.3.0
   * `github.com/hashicorp/go-hclog` v1.6.2 -> v1.6.3
   * `github.com/hashicorp/vault/api` v1.11.0 -> v1.12.2
   * `github.com/hashicorp/vault/sdk` v0.10.2 -> v0.12.0
   * `github.com/stretchr/testify` v1.8.4 -> v1.9.0
   * `golang.org/x/oauth2` v0.16.0 -> v0.19.0
   * `google.golang.org/api` v0.161.0 -> v0.172.0

PR #200, opened 2024-04-18
Update mocks

tomhjp:
#199 updated the SDK to a version that added an extra method to the `logical.SystemView` interface which broke the build for the mocks.


tvoran (approved):

tomhjp:
Thanks!

PR #201, opened 2024-04-18
Changelog++ for v0.16.3 release + bump go and net versions

tomhjp:
Preparing for v0.16.3 release which will also be released as a separate binary on releases.hashicorp.com for the first time.


.go-version
```diff
@@ -1 +1 @@
-1.20.7
+1.22.2
```
tomhjp, line 1: It needs checking separately, but I think we might be able to get rid of these files once we're consistently using the new go.mod toolchain directive.

swenson (approved):
LGTM

tomhjp:
Thanks!

PR #203, opened 2024-04-19
Remove gopkg.in/square/go-jose.v2 dep

tomhjp:
There is no fix for the [GO-2024-2631](https://pkg.go.dev/vuln/GO-2024-2631) vuln for `gopkg.in/square/go-jose.v2`, so now seems like a good time to remove it from the dependencies and standardise on the latest v4 version of the module.

Unfortunately it does mean we need to be more selective about the signature algorithms we allow when receiving a JWT during a login request. I think RS256 is the correct choice based on empirical evidence, but I do have concerns that there's no public commitment to that signature algorithm in Google's docs, so it's possible they could change this signature algorithm at any point.

References:
* Explainer for how JWTs relate to the auth method: https://developer.hashicorp.com/vault/docs/auth/gcp#generating-jwts
* PR that addressed the CVE and introduced the requirement for a limited set of supported signature algorithms: https://github.com/go-jose/go-jose/pull/74
* Google API docs for generating the JWT in question: https://cloud.google.com/iam/docs/reference/credentials/rest/v1/projects.serviceAccounts/signJwt

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,7 @@ IMPROVEMENTS:
    * `github.com/stretchr/testify` v1.8.4 -> v1.9.0
    * `golang.org/x/oauth2` v0.16.0 -> v0.19.0
    * `google.golang.org/api` v0.161.0 -> v0.172.0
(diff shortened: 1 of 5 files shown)


swenson (approved):
LGTM, though based on our discussions, I think it would be a good idea to support `HS256` and `ES256` as well.

tomhjp:
Thanks, added those 2 as well 👍 I also updated the target release version to 0.17.0 as that seems more appropriate with the change in accepted algorithms.

PR #220, opened 2024-11-20
backport/1.16.x: update go-jose and other dependencies

kpcraig:
This is part of our ongoing goal of getting all instances of go-jose dependency to >=2.6.3 (or >=3.0.3, or >= 4.0.1) in all code.

There were also some other dependencies flagged by the scan; those are updated here too. This process resulted in a toolchain bump (brought in by `x/net`, i think.

`.go-version` had to be bumped to avoid a vuln in the stdlib.

The mocks had to be updated due to the sdk bump (#200)

The `Makefile` had to be fixed (#201)

This should be version 0.16.3


.go-version
```diff
@@ -1 +1 @@
-1.20.7
+1.22.7
```
tvoran, line 1: Looks like 1.22.9 is [out now](https://go.dev/doc/devel/release#go1.22.minor)?

```suggestion
1.22.9
```

go.mod
```diff
@@ -1,66 +1,68 @@
 module github.com/hashicorp/vault-plugin-auth-gcp
 
-go 1.20
+go 1.22
+
+toolchain go1.23.3
```
tvoran, line 5: Probably should remove this since Vault 1.16.x is still on 1.22?

```suggestion
```

tvoran (approved):

PR #232, opened 2025-06-18
PCI Compliance VAULT-36018

tomcf-hcp:
PCI Compliance VAULT-36018


tvoran:
done in https://github.com/hashicorp/vault-plugin-auth-gcp/pull/234

PR #246, opened 2025-11-06
fix: cached credentials can be unusable if short lived

BaptisteLalanne:
The ctx used to create the cached credentials make the renewal of the token impossible blocking all authentifications still using the cached credentials.

The issue should only impact JWT expiration < 2 * cacheTime (2 * 30mn) - as httpclient is cached with the credentials and the credentials itself is cached. As mentioned [here](https://github.com/hashicorp/vault-plugin-auth-gcp/pull/63) the cache was initially meant for 60mn jwt but with the different ways to create credentials this is not always true specially for identity workload users

# Overview

### What is the change? 

A bug fix how on the credentials are created to allow auto-renewal.

### Why is the change needed?

The credentials are cached for 30mn. The creds cached context is tied to the request context that created it. If the cached token is expired, Google SDK will try to renew it doing http calls with that same context:
https://cs.opensource.google/go/x/oauth2/+/master:oauth2.go;l=244-260
https://cs.opensource.google/go/x/oauth2/+/master:token.go;l=171-183;drc=792c8776358f0c8689d84eef0d0c966937d560fb

The thing is, doing those calls to renew the token will fail if the context for the initial request that created the cache entry is canceled.

It will thus generate those errors for every `auth/gcp/login` calls:
```
oauth2/google: unable to generate access token: Post "https://iamcredentials.googleapis.com/v1/projects/-/serviceAccounts/xxx@yyyyy.iam.gserviceaccount.com:generateAccessToken": oauth2/google: invalid response from Secure Token Server: Post "https://sts.googleapis.com/v1/token": context canceled
```

Reloading the plugin with the exact same instance and credentials instantly fixes the issue

Same as for the http client that is cached, it should use context background instead of the request context to decouple the cache's context and the request hitting a cache miss:
https://github.com/hashicorp/vault-plugin-auth-gcp/blob/b8a1e7d17c16a68c2bb635c4bd07107b99b4ed9f/plugin/backend.go#L205

# Contributor Checklist

It's a bug fix so not sure if those below are relevant ⬇️ 
[My Docs PR Link](link)
[ ] Backwards compatible
[ ] Changelog entry added. See [Updating the Changelog](https://github.com/hashicorp/vault-plugin-auth-gcp/blob/main/README.md#updating-the-changelog).

## PCI review checklist

- [ ] I have documented a clear reason for, and description of, the change I am making.

- [ ] If applicable, I've documented a plan to revert these changes if they require more than reverting the pull request.

- [ ] If applicable, I've documented the impact of any changes to security controls.

  Examples of changes to security controls include using new access control methods, adding or removing logging pipelines, etc.

### DIFF
--- a/plugin/backend.go
+++ b/plugin/backend.go
@@ -224,44 +224,69 @@ func (b *GcpAuthBackend) clientOptions(ctx context.Context, s logical.Storage, e
 // for performance.
 func (b *GcpAuthBackend) credentials(ctx context.Context, s logical.Storage) (*google.Credentials, error) {
 	creds, err := b.cache.Fetch("credentials", cacheTime, func() (interface{}, error) {
-		b.Logger().Debug("loading credentials")
-
-		config, err := b.config(ctx, s)
-		if err != nil {
-			return nil, err
+		type result struct {
+			creds *google.Credentials
+			err   error
 		}
+		r := make(chan result, 1)
+		credsCtx, cancel := context.WithCancel(context.Background())
 
-		// Get creds from the config
-		credBytes, err := config.formatAndMarshalCredentials()
-		if err != nil {
-			return nil, fmt.Errorf("failed to marshal credential JSON: %w", err)
-		}
+		go func() {
+			b.Logger().Debug("loading credentials")
 
-		// If credentials were provided, use those. Otherwise fall back to the
-		// default application credentials.
-		var creds *google.Credentials
-		if len(credBytes) > 0 {
-			creds, err = google.CredentialsFromJSON(ctx, credBytes, iam.CloudPlatformScope)
+			var err error
+			defer func() {
+				if err != nil {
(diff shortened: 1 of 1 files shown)


pl-jankowskimichal (approved):

pl-jankowskimichal:
@jefferai any chance to get this reviewed and merged?

bayanoj330-dev (approved):

BaptisteLalanne:
Sadly this PR wasn't reviewed. Yet Hashiscorp did fix the issue in an equivalent PR https://github.com/hashicorp/vault-plugin-auth-gcp/pull/251 so I'm closing it.

The bug was reported and the same fix was proposed in october 2025 (https://github.com/hashicorp/vault-plugin-auth-gcp/commit/9c2f6da9c113346aa1bc568311d9c2422769a599)

PR #260, opened 2026-05-12
Bump vault sdk to v0.25.1 and api to v1.23.0

dmitryax:
Drop the indirect github.com/jackc/pgx/v4 dependency in favor of github.com/jackc/pgx/v5, which is brought in by vault sdk v0.25.1 (shipped with Vault server v2.0.0).


dmitryax:
@maithyton can you please take a look?

maithyton (approved):
Thank you, @dmitryax!

